PR #5, opened 2025-10-31
Add document for user journeys

rramkumar1:
Added the initial CUJs we have been focusing on. Please leave comments!

### DIFF
--- a/docs/user-journeys.md
+++ b/docs/user-journeys.md
@@ -1,0 +1,51 @@
+# Personas
+
+**AI Engineer**: A hands-on builder focused on the end-to-end development, deployment, and optimization of AI agents. They are distinct from ML Researchers and ML Engineers; AI Engineers are product-first, operating on the other side of the LLM Inference Serving API, and are not responsible for training, tuning, or deploying the models themselves.
+
+**Platform Engineer**: A builder and operator of the foundational platform (e.g. the provider of ingress/egress Gateways for the cluster).
+
+**AI Platform Engineer**: A builder and operator that leverages the foundational platform and builds layers on top that enable AI engineers to develop and deploy agents at scale. 
+
+**AI Security Engineer**: A specialist focused on designing safeguards to ensure AI agents operate safely and securely.
+
+**Application Developer**: A builder that is primarily focused on traditional APIs / applications but also surfaces functionality to agents using MCP.
+
+**Tool Developer**: A builder focused on developing MCP tools that can be leveraged by agents.
+
+# CUJs
+
+## Agent Identity 
+
+As an AI Engineer, I want to assign a unique, verifiable identity to my agent running in Kubernetes, so that gateways or external systems can securely authenticate it and make authorization decisions.
+
+## Protocol-Aware Authorization
+
+As an AI Platform Engineer, I want to:
+
+* Deny any traffic coming from Agents to MCP servers & other Agents by default
+
+* Allow agents to connect to specific, defined sets of MCP servers (e.g. "toolsets", "virtual service")
+
+* Allow agents to use specific tools
+
+* Allow agents to use specific tools from specific MCP servers
+
+* Control whether access to tools is read, write or both
+
+## Observability
+
+As an AI Engineer I want to:
+
+* Understand why my agent is getting denied when calling a certain tool
+
+As an AI Platform Engineer I want to:
+
+* Have an aggregated way of seeing failures/denials across the platform
+
+## Security
+
+As an AI Security Engineer I want to:
+
+* Develop MCP guardrails for pre-request filtering to prevent attacks such as prompt injection.
+
+* Develop MCP guardrails for post-response filtering to prevent data breaches.


docs/user-journeys.md
```diff
@@ -0,0 +1,27 @@
...
+
+**AI Platform Engineer**: A builder and operator of the foundational platform that enables AI engineers to develop and deploy agents at scale. 
+
+**Tool Developer**: A builder focused on developing MCP tools that can be leveraged by agents.
+
+# CUJs
+
+## Agent Identity 
+
+As an AI Engineer, I want to assign a unique, verifiable identity to my agent running in Kubernetes, so that gateways or external systems can securely authenticate it and make authorization decisions.
```
shaneutt: :+1: excellent

docs/user-journeys.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Protocol-Aware Authorization
+
+As an AI Platform Engineer, I want to:
+
+* Establish a zero-trust environment for traffic coming from agents
+
+* Allow agents to connect to specific MCP servers
```
shaneutt: I might use a bit stronger language here:

```suggestion
* Allow agents to connect to specific, defined sets of MCP servers (e.g. "toolsets", "virtual service")
```

Or whatever the latest language is for this.

docs/user-journeys.md
```diff
@@ -0,0 +1,27 @@
...
+
+* Allow agents to use specific tools
+
+* Allow agents to use specific tools from specific MCP servers
+
+* Control whether access to tools is read, write or both
```
shaneutt: :+1: to this being a goal and motivation, but given the current ecosystem I expect this to be interesting.

docs/user-journeys.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Agent Identity 
+
+As an AI Engineer, I want to assign a unique, verifiable identity to my agent running in Kubernetes, so that gateways or external systems can securely authenticate it and make authorization decisions.
+
+## Protocol-Aware Authorization
+
+As an AI Platform Engineer, I want to:
+
+* Establish a zero-trust environment for traffic coming from agents
```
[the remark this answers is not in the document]
david-martin:
```suggestion
* Deny any traffic coming from Agents to MCP servers & other Agents by default
```
Maybe something like this, combined with the more specific journies below?

docs/user-journeys.md
```diff
@@ -0,0 +1,27 @@
+# Personas
+
+**AI Engineer**: A hands-on builder focused on the end-to-end development, deployment, and optimization of AI agents. They are distinct from ML Researchers and ML Engineers; AI Engineers are product-first, operating on the other side of the LLM Inference Serving API, and are not responsible for training, tuning, or deploying the models themselves.
+
+**AI Platform Engineer**: A builder and operator of the foundational platform that enables AI engineers to develop and deploy agents at scale.
```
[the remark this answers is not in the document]
rramkumar1: Done. Added a new section called "Observability" and put those two there.

docs/user-journeys.md
```diff
@@ -0,0 +1,37 @@
+# Personas
+
+**AI Engineer**: A hands-on builder focused on the end-to-end development, deployment, and optimization of AI agents. They are distinct from ML Researchers and ML Engineers; AI Engineers are product-first, operating on the other side of the LLM Inference Serving API, and are not responsible for training, tuning, or deploying the models themselves.
+
+**AI Platform Engineer**: A builder and operator of the foundational platform that enables AI engineers to develop and deploy agents at scale. 
+
+**Tool Developer**: A builder focused on developing MCP tools that can be leveraged by agents.
+
+# CUJs
```
szuecs: As platform engineer (or gateway engineer), I would like to be able to route http traffic by header data.
As far as I know AI flavor is written in the json body and requires llm gateways to inspect the body to correctly route to the llm backend.

docs/user-journeys.md
```diff
@@ -0,0 +1,27 @@
+# Personas
+
+**AI Engineer**: A hands-on builder focused on the end-to-end development, deployment, and optimization of AI agents. They are distinct from ML Researchers and ML Engineers; AI Engineers are product-first, operating on the other side of the LLM Inference Serving API, and are not responsible for training, tuning, or deploying the models themselves.
+
+**AI Platform Engineer**: A builder and operator of the foundational platform that enables AI engineers to develop and deploy agents at scale. 
+
+**Tool Developer**: A builder focused on developing MCP tools that can be leveraged by agents.
+
```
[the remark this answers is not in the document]
shaneutt: I think it's important to have a decent overview in terms of stories up front, since the intention is usually that they drive everything else afterward, and then work that into your goals. This can be good communication for newcomers coming into the project as it starts too. However in the interest of time, it doesn't mean that the first iteration of the API has to achieve these goals yet.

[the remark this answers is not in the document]
rramkumar1: I added an AI Security Engineer persona and then an attached story specifically for guardrails.

docs/user-journeys.md
szuecs: Does it help or do I need to phrase it for adding it to the PR?

szuecs: There's also the routing latency that you add if the decision is in the body to which backends to route to.
So it's not only about policy.

Often you have a browser client which sends one tcp packet which contains the header and then sends the body in separate packets. The first packet will be used to connect and route in a distributed system. The callstack between multiple services can be deep.
The current openAI spec slows this down by a huge margin.

david-martin: +1 on routing latency being a concern @szuecs 

Thinking forward on this, I'm wondering what can we reasonably do in kubernetes to improve or standardise around this, and coming back from that, what are the user journeys.
Routing based on headers sounds like a reasonable thing a gateway should do.
The client needs to send that header, and the client conforms to a spec.

Are we talking about the OpenAI spec (and possibly also the MCP spec as that has some data that could be used for routing decisions, like tool name)? A2A as well?
Or focused on what the gateway should do if those specs evolve to have routing info in headers?

docs/user-journeys.md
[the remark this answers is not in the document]
shaneutt: Actually I think the distinction is important: The AI platform operator might be utilizing Gateways provided by another persona to enable networking, in the same way that it might be relying on another persona to deploy the cluster network, or provision load-balancer `Services`.

docs/user-journeys.md
rramkumar1: > route HTTP traffic by HTTP header to efficiently route to the right LLM or MCP Servers (MCP should send "model" in the header)

I understand the ask but this doesn't feel like it's in scope of this particular project.

> Provide authentication/authorization as a service for users and agents of LLM or MCP Servers

WDYM by authn/authz "as a service"?

LiorLieberman: @szuecs we meet tomorrow (Nov 6th) 8am pacific time, would you be able to attend?

PR #8, opened 2025-11-04
Propose Tool Authorization API

haiyanmeng:
This proposal defines authorization policies for tool access from AI agents running inside a Kubernetes cluster to MCP servers running in the Kubernetes cluster or outside of the Kubernetes cluster. By default, an AI agent can call `initialize`, `notifications/initialized` and `tools/list`. To enforce a "zero trust" security posture, a `tools/call` is denied unless it is allowed through the Tool Auth API described in this proposal.

The API introduces 2 new CRDs:

* `Backend`: describes a backend in agentic networking
* `AccessPolicy`: describes who can access what (the permissions/grants) in relation to the agentic networking backends

### DIFF
--- a/docs/proposals/0008-ToolAuthAPI.md
+++ b/docs/proposals/0008-ToolAuthAPI.md
@@ -1,3 +1,7 @@
+Date: 17th November 2025
+Authors: haiyanmeng, guicassolato
+Status: Provisional
+
 # Tool Authorization in Agentic Networking
 
 This proposal defines authorization policies for tool access from AI agents running inside a Kubernetes cluster to MCP servers running in the Kubernetes cluster or outside of the Kubernetes cluster. By default, an AI agent can call initialize, notifications/initialized and tools/list. To enforce a "zero trust" security posture, a tools/call is denied unless it is allowed through the Tool Auth API described in this proposal.


docs/proposals/0008-ToolAuthAPI.md
```diff
@@ -0,0 +1,459 @@
+# Tool Authorization in Agentic Networking
+
+This proposal defines authorization policies for tool access from AI agents running inside a Kubernetes cluster to MCP servers running in the Kubernetes cluster or outside of the Kubernetes cluster. By default, an AI agent can call initialize, notifications/initialized and tools/list. To enforce a "zero trust" security posture, a tools/call is denied unless it is allowed through the Tool Auth API described in this proposal.
+
+# Non-Goals
+
+The authentication of MCP tool access is not within the scope of this proposal, and will be explored separately in the future.
+
+# Use Cases & Motivation
```
haiyanmeng: Reference the personas and user journeys described in https://github.com/kubernetes-sigs/kube-agentic-networking/pull/5 after it is merged.

docs/proposals/0008-ToolAuthAPI.md
```diff
@@ -0,0 +1,459 @@
...
+// When the action is ALLOW,
+//   - requests from Source are permitted to access the listed Tools.
+//   - If the tool list is empty, the rule denies access to all tools from Source.
+type AuthRule struct {
+	// Source specifies the source of the request.
+	// +required
+	Source Source `json:"source"`
+	// Tools specifies a list of tools.
+	// +optional
+	Tools []string `json:"tools,omitempty"`
```
guicassolato: I've shared my thoughts about this proposed API in the [design doc](https://docs.google.com/document/d/1QLJ_X99VfTgRCzwAaoBzsBq_u4anxT2N5kjlCSJNROc/edit?disco=AAABtS33dsA) already, but now feel like this is the place to put it on record again.

I strongly believe this may be going in the wrong direction. This API that consists on listing resources (today tools, but soon likely resources and prompts too) and principals all in an AuthPolicy CR have a few problems that IMO should not be overlooked:
1. it doesn't scale well for large number of resources and principals;
2. it copies very specific existing approaches, being heavily influenced by how one implementation in particular would probably support it underneath, rather than focusing UX first;
3. it oversteps on Kubernetes RBAC as the mean for storing authorisation data.

docs/proposals/0008-ToolAuthAPI.md
```diff
@@ -0,0 +1,459 @@
...
+
+## User Journeys
+
+#### Agent Identity 
+
+As an AI Engineer, I want to assign a unique, verifiable identity to my agent running in Kubernetes, so that gateways or external systems can securely authenticate it and make authorization decisions.
+
+#### Protocol-Aware Authorization for MCP Tools
+
+As an AI Engineer, I want to create authorization policies to specify which individual tools (e.g., getWeather, sendEmail) my agent is permitted to call on an allow-listed MCP server, so that I can enforce least-privilege access at the specific tool-function level, not just the network endpoint.
```
david-martin: Following on from a comment discussion in the original google doc, 
how do you feel about a user journey about filtering the list of tools returned from a tool list so that tokens and time are not wasted trying to call a tool that a user doesn't have access to?

docs/proposals/0008-ToolAuthAPI.md
```diff
@@ -0,0 +1,459 @@
...
+
+// BackendSpec defines the desired state of Backend.
+type BackendSpec struct {
+	// Type specifies the type of the backend.
+	// Currently, only "MCP" is supported.
+	// +required
+	Type *BackendType `json:"type"`
+	// MCP defines a MCP backend.
+	// +optional
+	MCP MCPBackend `json:"mcp,omitempty"`
```
shaneutt: This `Backend` API is representative of [egress gateways](https://github.com/kubernetes-sigs/wg-ai-gateway/blob/main/proposals/10-egress-gateways.md), a use case that transcends agentic, and AI in general. Having `MCPBackend` makes _this_ more specifically an agentic AI API.

I expect this capability belongs in Gateway API itself. If we move forward with this API here, how do we see threading the needle with all the other use cases and stakeholders for this?

haiyanmeng: This is discussed in https://kubernetes.slack.com/archives/C09P6KS6EQZ/p1762448311074899

usize: FWIW I had planned to base the `Backend` in [this proposal](https://github.com/kubernetes-sigs/wg-ai-gateway/pull/16) on what you have here in order to avoid repeating work. A GEP would make my life a bit easier in that regard.

docs/proposals/0008-ToolAuthAPI.md
```diff
@@ -0,0 +1,459 @@
...
+
+*Tool Developer:* A builder focused on developing MCP tools that can be leveraged by agents.
+
+*AI Security Engineer:* A specialist focused on designing automated safeguards to ensure AI agents operate safely and securely, while sometimes also wearing multiple hats (e.g., policy architect, risk manager, and compliance advisor) ensuring security is integrated across all aspects of AI development and deployment.
+
+## User Journeys
+
+#### Agent Identity 
+
+As an AI Engineer, I want to assign a unique, verifiable identity to my agent running in Kubernetes, so that gateways or external systems can securely authenticate it and make authorization decisions.
```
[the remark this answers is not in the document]
guicassolato: I see "assigning a unique, verifiable identity to an agent" as incompatible with authentication being a non-goal.

docs/proposals/0008-ToolAuthAPI.md
```diff
@@ -0,0 +1,657 @@
...
+
+# API
+
+The API introduces 2 new CRDs:
+- `Backend`: describes a backend in agentic networking
+- `AccessPolicy`: describes who can access what (the permissions/grants) in relation to the agentic networking backends
+
+The CRD names may change depending on the OSS feedback.
+
+> **_NOTE:_** The API does not cover identity extraction, or request authentication. We will cover them in a follow-up Pull Request.
```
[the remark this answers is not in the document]
LiorLieberman: >This creates scalability challenges with:
I dont think we should be bothered with scalability at this stage. We are in a prototyping phase. I would like to understand more the potential conflicts argument though. 

The AuthScheme, while robust is very opinionated way of doing something, and its complex. I do want to have more discussion on that in a wider forum. 

I think its a win that we have some consensus on "Backend" and "AccessPolicy" (and thanks for the awesome changes you suggested) but my real concern is that we start with something very complex to begin with. I'd rather tackle it one by one. 

Backed is already a new type folks needs to be doing, and AccessPolicy is another one. 

Anyway, I guess the summary of what I say is, I do understand AuthScheme place, but I would love to have that as a fast follow thing to discuss and see exactly the user flows we are covering and how. We spent a bit amount of time discussing Backend and AccessPolicy (or formerly the ToolAuthPolicy) to gather the feedback but we havent done this for AuthScheme. And I do think its important to allow a UX where users **dont need** an AuthScheme, rather they can use it (or some other mechanism, TBD based on discussions+feedback) to customize/enhance what they need.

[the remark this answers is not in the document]
Verolop: 👋 I've been lurking for a while with some ideas, but I just carved out some time to read the entire proposal and comments, so apologies for popping in randomly!

As a quick middle-ground proposal for the authN question: instead of defining the whole extraction flow in this PR, we could just define the canonical identity formats that AccessPolicy should match, similar to how fine-grained authorization systems like Zanzibar handle identities from different sources.

e.g., these three can all refer to the same workload:

- `spiffe://cluster.local/ns/default/sa/agent`

- `{ namespace: "default", name: "agent" }`

- `{ iss: "https://issuer.example.com", sub: "agent@default" }`

which can be normalised to something like: `sa:default/agent`

So we could define formats such as `sa:<ns>/<name>`, `spiffe:<uri>`, `oidc:<issuer>/<sub>`, `external:<id>`, and leave the details of how those identities get extracted to a follow-up proposal. Without this, two implementations could evaluate AccessPolicy against different identity strings for the same principal, causing potential consistency violations and policy drift.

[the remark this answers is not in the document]
guicassolato: Please let's discuss the `AuthScheme` in https://github.com/haiyanmeng/kube-agentic-networking/pull/1. I'm very interested in your points regarding it being opinionated and complex @LiorLieberman.

docs/proposals/0008-ToolAuthAPI.md
yuzisun: In agentic network we have three primary backends: LLMs(cloud or on prem hosted), MCP and Agent as an agent can call LLMs, tools or other agents. In envoy ecosystem, we have alpha version of `Backend` implemented in envoy gateway project and in envoy ai gateway project we introduced `AIServiceBackend` for LLM egress cases. I think there are two options:
1) Add a generic `Backend` in Gateway API and here we define `AgenticBackend` to have a reference to `Backend` 
2) Composite backend like the way defined here, but allow defining MCP, LLM, Agent specific backend fields.

david-martin: I think prototyping and iterating on a single CRD with a type field would be quicker than introducing additional CRDs at this stage.
So, 2 is my recommendation.

docs/proposals/0008-ToolAuthAPI.md
```diff
@@ -0,0 +1,459 @@
+# Tool Authorization in Agentic Networking
+
+This proposal defines authorization policies for tool access from AI agents running inside a Kubernetes cluster to MCP servers running in the Kubernetes cluster or outside of the Kubernetes cluster. By default, an AI agent can call initialize, notifications/initialized and tools/list. To enforce a "zero trust" security posture, a tools/call is denied unless it is allowed through the Tool Auth API described in this proposal.
```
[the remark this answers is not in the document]
robscott: I think the same API can likely be used for traffic entering a cluster, within a cluster, or leaving a cluster. With that said, within a cluster and leaving a cluster seem to be more common use cases at this point, and both can reuse the same way of expressing the client's identity, likely tied to a Kubernetes ServiceAccount. For traffic entering a cluster, the ways to express a client identity seem more open ended and therefore complex, so I'd suggest leaving this for a future iteration of the proposal.

docs/proposals/0008-ToolAuthAPI.md
LiorLieberman: Yes I am also plus 1 to 2 as well. I think its fine to start with MCP as a field here though and add more BackendTypes later -- @david-martin @yuzisun agreed?

docs/proposals/0008-ToolAuthAPI.md
LiorLieberman: These concerns are resolved thanks to your improvements to the original PR (which were partially brought back here). AuthScheme seem to solve all of it thought but I think it needs more discussion. 

I would like to have a PR with AuthScheme proposal only and get more comments on that specifically. I think its a great API to have, but I know there were some concerns about the specificity and whether it will work for other implementations. 

I should also say that since we are in a prototyping phase, it will surely be ok to prototype with AuthScheme, but I would like to iterate on this.

docs/proposals/0008-ToolAuthAPI.md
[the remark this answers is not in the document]
LiorLieberman: actually I think that was a duplicate of this comment https://github.com/kubernetes-sigs/kube-agentic-networking/pull/8#discussion_r2533576041

I think I am plus one with your suggestion there. Resolving this comment and re-opened the one I linked to make sure we have that suggestion incorporated

docs/proposals/0008-ToolAuthAPI.md
```diff
@@ -0,0 +1,665 @@
...
+
+# 🚫🚫 **STOP – EXPERIMENTAL API** 🚫🚫
+**Do NOT use this in production.**
+
+This API is currently **experimental** and is subject to change without prior notice. We recommend against using it in production environments until a stable version is released.
```
guicassolato:
```suggestion
**Do NOT implement. Do NOT use in production.**

This API is **experimental** and subject to change without prior notice. Vendors and integrators should not implement or rely on it, and it must not be enabled in production environments until a stable version is released.
```

I'd even go with "provisional" instead of "experimental". We want to highlight this is just a proposal at this point, still pending on prototyping and reiteration before vendors start adding it to their products (even as an experimental feature).

guicassolato (approved):
Since I originally blocked this PR on technical grounds, some of the time spent working toward unblocking it – KubeCon NA excluded – has ended up being focused on process-related aspects rather than the proposal itself. We now seem to have an [appropriate place](https://github.com/kubernetes-sigs/kube-agentic-networking/pull/10) to continue those process discussions.

Although not all technical concerns have been addressed at this stage – whether the ones I raised or the many others kindly contributed by the community – I'm going to unblock the PR. I trust that any remaining technical issues will surface again during the prototyping phase, where they can be discussed concretely in the context of code.

I will also close https://github.com/haiyanmeng/kube-agentic-networking/pull/1 and reopen it as a separate proposal targeting `main` branch.

Thank you, @haiyanmeng, for your patience and for addressing as many comments as you could, especially amid some heated discussions.

I'm leaving just a couple of new comments that I strongly recommend addressing, but I will not let them cause further delays.

docs/proposals/0008-ToolAuthAPI.md
david-martin: +1 to 'Provisional', with an understanding it aligns to something like https://gateway-api.sigs.k8s.io/geps/overview/#gep-states

> Provisional: The goals described by this GEP have consensus but implementation details have not been agreed to yet.

LiorLieberman (approved):
Thank you everyone for your hard work on this. Hopefully this gives us a good start for iteration. 

@Verolop @maia-iyer you had some good feedback on the PR here and in https://github.com/haiyanmeng/kube-agentic-networking/pull/1

Can you please collect the feedback again in an issue? this will be the best place to track it, and we dont want to miss this feedback

/lgtm

PR #33, opened 2025-12-19
Observability: OTel semantic conventions for agentic systems

david-martin:
**What type of PR is this?**

/kind documentation

Defines tracing schemas for agentic systems by following OpenTelemetry semantic conventions, with proposed extensions for permission enforcement outcomes.

### DIFF
--- a/docs/proposals/0033-Observability.md
+++ b/docs/proposals/0033-Observability.md
@@ -99,17 +99,26 @@ Span: gen_ai.agent.chat                         [span_id: 3d4e5f6a]
       │
       │ POST /mcp
       │ Headers:
-      │   traceparent: 00-f5a9d214e6b8c7a9d1e2f3a4b5c6d7e8-5e6f7a8b9c1d2e3f-01
+      │   traceparent: 00-f5a9d214e6b8c7a9d1e2f3a4b5c6d7e8-4e5f6a7b-01
       │   Authorization: Bearer eyJhbGc...
       │   ...
-      │ tool: delete_customer_data
+      │ Body (JSON-RPC):
+      │ {
+      │   "jsonrpc": "2.0",
+      │   "method": "tools/call",
+      │   "params": {
+      │     "name": "delete_customer_data",
+      │     "arguments": { "customer_id": "12345" }
+      │   },
+      │   "id": 1
+      │ }
       ▼
 
 Span: mcp.gateway.request                       [span_id: 5e6f7a8b]
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 17ms ━━━━━━━━━━━━
 ├─ trace_id: f5a9d214e6b8c7a9d1e2f3a4b5c6d7e8 ← Same trace_id
 ├─ span.kind: SERVER
-├─ parent_span_id: 3d4e5f6a
+├─ parent_span_id: 4e5f6a7b
 ├─ http.response.status_code: 403
 │
 ├─ gen_ai.operation.name: execute_tool
@@ -212,9 +221,11 @@ Span: gen_ai.agent.chat                         [span_id: 3d4e5f6a]
       │
       │ POST /mcp
       │ Headers:
-      │   traceparent: 00-c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8-9c1d2e3f4a5b6c7d-01
+      │   traceparent: 00-c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8-4e5f6a7b-01
       │   ...
+      │ Body (JSON-RPC):
       │ {
+      │   "jsonrpc": "2.0",
       │   "method": "tools/call",
       │   "params": {
       │     "name": "send_email",
@@ -222,7 +233,8 @@ Span: gen_ai.agent.chat                         [span_id: 3d4e5f6a]
       │       "to": "bob@example.com",
       │       "body": "SSN: 123-45-6789, CC: 4532-1234-5678-9010"
       │     }
-      │   }
+      │   },
+      │   "id": 2
       │ }
       ▼
 
@@ -230,7 +242,7 @@ Span: mcp.gateway.request                       [span_id: 5e6f7a8b]
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 30ms ━━━━━━━━━━━━
 ├─ trace_id: c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8 ← Same trace_id
 ├─ span.kind: SERVER
-├─ parent_span_id: 3d4e5f6a
+├─ parent_span_id: 4e5f6a7b
 ├─ http.response.status_code: 400
 │
 ├─ gen_ai.operation.name: execute_tool


docs/proposals/0033-Observabilty.md
```diff
@@ -0,0 +1,61 @@
...
+- **tool.session_id**: Session identifier for the tool invocation
+- **permission.policy**: AccessPolicy evaluated for the operation
+- **permission.rule**: Specific rule within the AccessPolicy that caused success/failure
+- **permission.result**: Whether the permission check passed or failed
+
+### Identity Propagation Options
+
+There are two approaches for propagating user and agent identity, which can be used individually or in combination:
+
+1. **Header-based propagation**: Include user and agent identifiers in custom headers (e.g., `X-User-ID`, `X-Agent-ID`) alongside the `traceparent` header. This makes each span independently queryable but adds overhead to every request.
```
gkhom: This would require all agents/tools to propagate these additional headers to downstream calls. Baggage (W3C Baggage) should be considered: https://w3c.github.io/baggage/

docs/proposals/0033-Observabilty.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Problem space
+
+Users of Agentic systems can set goals for an Agent to solve. For example, "Generate a sales report for Q4 and share it with the leadership team". Agents use LLMs and tools in a loop to reach the goal. This will require tools like 'database query', 'report generation', and 'email/messaging'. LLM access will be required for things like analyzing sales data patterns, formatting the report, and composing the distribution message.
+
+If something goes wrong while the Agent is solving the goal, such as the Agent attempting to access restricted financial data beyond its permitted scope, or trying to share the report with unauthorized recipients, it's necessary to understand where the Agent went wrong and why.
+
+This requires solving several observability challenges:
+- How to trace an agent's entire execution flow from the initial user request through to completion
+- How to capture which LLM calls were made, what prompts were sent, and what responses were received
```
gkhom: Capturing prompts and responses seems risky. They often contain PII or secrets. Perhaps the proposal should contain guidance on data masking? Standardized attributes should define what is "safe" to log versus what must be redacted or considered optional.

david-martin: Perhaps some wording about those risks, and proposing an 'opt in' approach?

Data masking could be an optional 3rd, less risky approach.

docs/proposals/0033-Observabilty.md
```diff
@@ -0,0 +1,61 @@
...
+- Standardized logging format that includes user identity, agent identity, and permission check results
+- When a tool request fails due to an AccessPolicy, the audit trail should show the specific AccessPolicy and rule that caused the failure
+
+### Distributed Tracing Approach
+
+Leverage distributed tracing standards (W3C Trace Context) to track agent execution flows. The W3C `traceparent` header provides:
+- **Trace ID**: A unique identifier for the entire agent session (from initial user goal to completion), propagated across all components
+- **Span ID**: A unique identifier for each individual operation (e.g., a single LLM call or tool invocation)
+
+Each component (agent runtime, LLM providers, MCP servers) propagates the trace context via standard headers and creates spans for its operations. All spans share the same trace ID, allowing the full flow to be reconstructed.
```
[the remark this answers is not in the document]
david-martin: There has been some discussion about adding trace information to `request.params._meta`

* https://github.com/modelcontextprotocol/modelcontextprotocol/issues/246
* https://github.com/modelcontextprotocol/modelcontextprotocol/pull/414

With one of the more recent comment on that [2nd issue saying](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/414#issuecomment-2908208066):

> ...there are enough artifacts here and there to suggest request.params._meta is already in use by the spec...

I've also just come across https://github.com/open-telemetry/semantic-conventions/pull/2083,
which I need to read through

docs/proposals/0033-Observabilty.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Standardized Trace Attributes
+
+Extend standard tracing with agent-specific attributes on each span:
+
+- **agent.id**: Identifier for the agent executing the operation
+- **user.id**: Identifier for the user who delegated authority to the agent
```
rubambiza: Should this be a pseudo-userid? Just like browsers in W3C cannot freely attach headers for privacy reasons, I think the same principle applies here. For example, I would not want downstream trackers to trace who I am when I delegate authority to my agent to buy something in an online cart.

Perhaps there is a larger discussion to be had about who/what _generate(s)_ the trace parents/states vs which backends are allowed to _analyze_ them. I could be okay with this being known within my k8s cluster/namespace, but I wouldn't want that to be the case to the outside world.

docs/proposals/0033-Observabilty.md
```diff
@@ -0,0 +1,140 @@
...
+
+Follow [OpenTelemetry MCP semantic conventions](https://github.com/open-telemetry/semantic-conventions/pull/2083) such as:
+- `mcp.method.name`: Name of the MCP method being invoked (e.g., "tools/call")
+- `mcp.session.id`: Session identifier for the MCP connection
+- `gen_ai.operation.name`: The operation being performed (e.g., "execute_tool")
+- `gen_ai.tool.name`: Name of the tool utilized by the agent. (e.g. "Flights")
+
+**Permission and AccessPolicy attributes (proposed):**
+
+For permission checks and AccessPolicy enforcement, we propose the following span attributes to capture authorization decisions:
```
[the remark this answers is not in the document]
david-martin: @gkhom do you think the overlap is sufficient to possibly reference them instead of defining new attributes?

There are 8 attributes, of which I think only the policy name could align (with security_rule.name).
The policy rule, and result (if the policy rule as allowed or denied) don't seem to have a natural mapping.

docs/proposals/0033-Observability.md
```diff
@@ -0,0 +1,304 @@
...
+- `mcp.method.name`: Name of the MCP method being invoked (e.g., "tools/call")
+- `mcp.session.id`: Session identifier for the MCP connection
+- `gen_ai.operation.name`: The operation being performed (e.g., "execute_tool")
+- `gen_ai.tool.name`: Name of the tool utilized by the agent. (e.g. "Flights")
+
+*Policy enforcement spans:*
+
+These can encompass spans for permission checks and AccessPolicy enforcement and can be generalized to other guardrails checks.
+
+Follow [OpenTelemetry security rule semantic conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/security-rule/) such as:
```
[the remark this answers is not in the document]
david-martin: Coming back to this, with a better understanding.

What about emitting a span for each rule evaluation, just up to the point of allowing?

e.g.

```
  Span: mcp.authorization                        [span_id: 5e6f7a8b]
  ├─ security_rule.ruleset.name: crm_data_access_policy
  ├─ event.action: allow
  ├─ event.outcome: success
  │
  ├─► Span: mcp.authorization.rule              [span_id: 6f7a8b9c]
  │   ├─ security_rule.name: rule-0
  │   ├─ security_rule.match: true
  │   └─ event.action: deny
  │
  ├─► Span: mcp.authorization.rule              [span_id: 7a8b9c0d]
  │   ├─ security_rule.name: rule-1
  │   ├─ security_rule.match: false
  │   └─ event.action: deny
  │
  └─► Span: mcp.authorization.rule              [span_id: 8b9c0d1e]
      ├─ security_rule.name: rule-2
      ├─ security_rule.match: false
      └─ event.action: allow
```

In this case, we're saying it doesn't matter if there are other rules after rule-2 that do allow a request, so we stop at that point with the allowing rule (why go further).
Am I missing something here, like for how `allow foo || bar and allow foo || baz` would be specified in an array of rules?

cc @evaline-ju @rubambiza

docs/proposals/0033-Observability.md
```diff
@@ -0,0 +1,288 @@
...
+These can be considered for inclusion in an existing or new OpenTelemetry semantic convention registry.
+
+### Per-Rule Evaluation Spans
+
+For authorization and guardrail checks, emit an `mcp.authorization.rule` or `mcp.guardrail.rule` child span for each rule evaluated within the parent span. Each child span records:
+- `security_rule.name` — the rule identifier
+- `security_rule.match` — whether the rule matched (`true`/`false`)
+- `event.action` — the action the rule would take (`allow`/`deny`)
+
+Evaluation stops at the determining rule. For authorization, if no allow rule matches, emit an explicit `default-deny` rule span (`security_rule.name: default-deny`, `security_rule.match: true`, `event.action: deny`) to make the implicit default-deny behavior visible in traces.
```
rubambiza: Just trying to understand: If the eval stops at the determining rule, does this mean the emission of default-deny will always happen at the end and `security.match: false` could never be exercised?

docs/proposals/0033-Observability.md
```diff
@@ -0,0 +1,288 @@
...
+        ├─ trace_id: f5a9d214e6b8c7a9d1e2f3a4b5c6d7e8  ← Same trace_id
+        ├─ span.kind: INTERNAL
+        ├─ parent_span_id: 3d4e5f6a
+        │
+        ├─ gen_ai.tool.name: delete_customer_data
+        └─ status: OK
+      │
+      │ POST /mcp
+      │ Headers:
+      │   traceparent: 00-f5a9d214e6b8c7a9d1e2f3a4b5c6d7e8-5e6f7a8b9c1d2e3f-01
```
rubambiza: Where is the `5e6f7a8b9c1d2e3f` part generated? 👀 It is partially addressed as the parent of `7a8b9c1d`, but I did not see where it was generated.

docs/proposals/0033-Observability.md
```diff
@@ -0,0 +1,288 @@
...
+      │   "method": "tools/call",
+      │   "params": {
+      │     "name": "send_email",
+      │     "arguments": {
+      │       "to": "bob@example.com",
+      │       "body": "SSN: 123-45-6789, CC: 4532-1234-5678-9010"
+      │     }
+      │   }
+      │ }
+      ▼
```
rubambiza: Is this the point where the traffic hits a k8s cluster?

rubambiza (commented):
Mostly clarifying questions, not sure if they may require any changes. They certainly do not need to block merging.

LiorLieberman:
@david-martin anything else thats blocking you from getting this in?

PR #69, opened 2026-02-09
TelemetryPolicy proposal

gkhom:
**What type of PR is this?**

/kind documentation

This PR contains a proposal for a new `TelemetryPolicy` API. This K8s API aims to standardize how users enable and configure telemetry (metrics, logs, traces) across different data plane implementations, replacing vendor-specific CRDs with a unified, portable spec.

Fixes #

```release-note
* TelemetryPolicy specification
```


haiyanmeng:
> [CLA Not Signed](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/20677499/1082323650/69/#/?version=2)
> 
> * [❌](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/20677499/1082323650/69/#/?version=2) - login: @gkhom / name: Georgi Khomeriki . The commit ([6bf817c](https://github.com/kubernetes-sigs/kube-agentic-networking/commit/6bf817c754c910ea72ea8c096bb44d9b1fb1c3cf), [972f464](https://github.com/kubernetes-sigs/kube-agentic-networking/commit/972f4640d330248cfa79f51398b4b84ab5fe0376)) is not authorized under a signed CLA. [Please click here to be authorized](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/20677499/1082323650/69/#/?version=2). For further assistance with EasyCLA, [please submit a support request ticket](https://jira.linuxfoundation.org/servicedesk/customer/portal/4).

@gkhom , can you fix this?

rubambiza (commented):
@gkhom Thanks for kicking this off. I left some clarifying questions and proposing a minor change. The bigger part of the review is to surface work already being proposed/done in the llm-d community with regard to tracing and whether it is applicable to our objectives.

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,49 @@
...
+Status: draft<br/>
+
+# TelemetryPolicy
+A Kubernetes API for Gateway/Mesh Observability
+
+## Summary
+This proposal introduces the `TelemetryPolicy`, a direct policy attachment designed to configure observability signals (metrics, logs, traces) 
+for Gateway API resources (via `Gateway` attachment) and Service Mesh resources (via `namespace` attachment).
+
+This K8s API standardizes how users enable and configure telemetry across different data plane implementations, replacing vendor-specific CRDs
```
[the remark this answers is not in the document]
gkhom: Examples of such CRDs are: 
* Istio's Telemetry CRD
* Envoy Gateway's EnvoyProxy and EnvoyGateway CRDs
* Kong's MeshMetrics/MeshTrace/MeshAccessLog
* Kuadrant's TelemetryPolicy

I intend to write a section that compares such existing APIs and the proposed TelemetryPolicy in the eventual proposal.

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,49 @@
...
+Existing telemetry policies are unaware of the Generative AI semantic conventions. They see an opaque TCP stream or HTTP POST. Without a standardized API to 
+configure the extraction and export of these attributes, the “Agentic Mesh” will remain a black box, increasing governance and cost control challenges.
+
+## Design Objectives
+
+To address these challenges, the `TelemetryPolicy` proposal targets four core objectives:
+
+1. **Standardization**: A single API for Gateway and Mesh to configure Access Logging, Metrics generation, and Tracing propagation.
+2. **GEP-713 Compliance**: Support `targetRef` attachment to `Gateway` and `Namespace`. The latter covers Mesh use-cases.
+3. **Agentic Support**: Enable the capture of OpenTelemetry GenAI Semantic Conventions and support the requirements of PR #33.
```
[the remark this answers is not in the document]
rubambiza: No, I did not see any CRD definitions. I'll keep this thread in mind as the definitions become more concrete.

david-martin:
I'm generally in favour of this proposal.
/approve

Perhaps more relevant when it comes to the specification, it would be good to know more about the current 'state of the art' in this space.

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,194 @@
...
+4. **Protocol Agnostic**: Support OpenTelemetry as the primary data model while allowing vendor-specific extensions.
+
+## The TelemetryPolicy Specification
+
+We propose the `TelemetryPolicy` as a direct policy attachment in the `gateway.networking.k8s.io` API group.
```
LiorLieberman:
```suggestion
We propose the `TelemetryPolicy` as a direct policy attachment in the `gateway.networking.k8s.io` API group. See [GEP-713](https://gateway-api.sigs.k8s.io/geps/gep-713/#classes-of-policies) for more information on Direct attachment.
```

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,194 @@
...
+  # GEP-713 Attachment
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: Gateway
+    name: my-gateway
+  
+  # 1. Tracing Configuration
+  tracing:
+    provider:
+      type: OTLP # or implementation-specific
```
LiorLieberman: Is type a specific go type? would be useful to define something like

```go
type TracingProvider string

const (
  // OTLPTracingProvider is used to ....

  OTLPTracingProvider TracingProvider = "OTLP"
)

```

And then an implementation specific would probably be a vendor-prefixed thing like `<foo.io/some-provider-name>`

Would probably be useful to have the go types as part of this PR like other GEPs in Gateway API

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,194 @@
...
+      endpoint: "otel-collector.monitoring.svc:4317"
+    samplingRate: 
+      percent: 5
```
LiorLieberman: would benefit from a go struct. We had loads of discussion on how to do percentages in gateway api. I could not find the long thread, I think it was in a meeting. This is the best thing I could fine https://github.com/kubernetes-sigs/gateway-api/pull/3178 but maybe @robscott has something more concrete

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,194 @@
...
+      - B3
+    customSpanAttributes:
+      - attributeName: "env"
+        literalValue: "production"
+
+  # 2. Metrics Configuration
+  metrics:
+    enable: true
+    provider:
+      type: Prometheus
```
LiorLieberman: same comment on go struct and types

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,194 @@
...
+    overrides:
+      - name: "request_count"
```
LiorLieberman: would be nice to have an example of a full metric name

LiorLieberman (commented):
took an initial pass, Thanks @gkhom !

docs/proposals/0069-TelemetryPolicy.md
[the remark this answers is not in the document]
gkhom: I'm leaning towards one resource for all. The argument for splitting them might be that different personas are involved in configuring the different aspects of observability. In practice, I think that the persona that configures metrics, likely also configures tracing and access logs. So to avoid complicating the API with three additional resources, it seems worthwhile to put all of it in a single resource.

docs/proposals/0069-TelemetryPolicy.md
gkhom: I've added Go types in the "Detailed resource description" section. I'm not sure that we need to include specific implementation providers as part of the API spec.

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,194 @@
...
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: Gateway
+    name: my-gateway
+  
+  # 1. Tracing Configuration
+  tracing:
+    provider:
+      type: OTLP # or implementation-specific
+      endpoint: "otel-collector.monitoring.svc:4317"
```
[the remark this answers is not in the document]
gkhom: Yes, in this example it's the URL of the OTLP endpoint.

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,194 @@
...
+      - "response_code"
+      - "x-token-usage"
+```
+
+### Policy Attachment
+
+Following [GEP-713](https://gateway-api.sigs.k8s.io/geps/gep-713/), the `TelemetryPolicy` supports the following attachments:
+
+1. **Gateway (Instance Scope)**: Configures the telemetry for a specific `Gateway`.
+2. **Namespace (Mesh Scope)**: Configures the telemetry for all mesh proxies (sidecar proxy / node proxy / etc.) in that namespace.
```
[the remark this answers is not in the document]
gkhom: My current thinking is that targeting a namespace implies a "mesh" target i.e., it would include all proxies in the namespace except gateways. I think it's reasonable to assume that different observability configurations may be desired/applicable to Gateway use-cases (north/south) compared to Mesh use-cases (east/west). That's why I think it's better to avoid making namespace a target that captures both.

docs/proposals/0069-TelemetryPolicy.md
gkhom: Updated the example and added comment in the Go structs.

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,295 @@
...
+    overrides:
+      - name: "gateway.networking.k8s.io/http/request_count"
+        type: Counter
+        dimensions: # Custom labels/dimensions
+          - key: "model_id"
+            fromHeader: "x-model-id" # Crucial for Agentic workloads
+
+  # 3. Access Logging
+  accessLogs:
+    enable: true
```
evaline-ju:
```suggestion
    enabled: true
```

evaline-ju (commented):
Generally looking good to me - a couple lingering questions

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,295 @@
...
+
+#### Metrics
+
+* **Granularity**: Users can enable/disable specific metric families.
+* **Dimensions**: The API supports "overrides" (similar to [OpenTelemetry Views](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#view)) where users can add or remove dimensions (labels/attributes) to reduce cardinality or increase visibility.
+
+#### Logging
+
+* **Flexible Formatting**: Supports both JSON and text formats for compatibility with standard log aggregation stacks.
+* **Smart Filtering**: Reduces noise and cost via CEL-based filtering, allowing logs to be generated only for specific events (e.g., 5xx errors, high latency, or critical paths).
```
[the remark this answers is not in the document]
gkhom: Great point! I completely forgot to add CEL matching to `MatchCondition`. Updated the proto and example.

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,317 @@
...
+      - attributeName: "env"
+        literalValue: "production"
+
+  # 2. Metrics Configuration
+  metrics:
+    enabled: true
+    provider:
+      type: Prometheus
+    overrides:
+      - name: "gateway.networking.k8s.io/http/request_count"
```
guicassolato: Are these metric names standardised somehow or the alignment between who owns the policy and who consumes the metrics is something we expect to happen but out of scope of the proposal?

guicassolato (commented):
I think overall this is a strong proposal with a compelling use case and well-designed API.

My main point of concern is whether it has enough to make it specific to the agentic use case or could otherwise better server upstream Gateway API as a whole.

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,317 @@
...
+    // Conditions for logging, allowing filtering to specific paths or events.
+    Matches []MatchCondition `json:"matches,omitempty"`
+
+    // A list of specific fields or headers to include in the logs.
+    Fields []string `json:"fields,omitempty"`
+}
+
+type MatchCondition struct {
+    // Path allows filtering to specific paths.
+    Path string `json:"path,omitempty"`
```
howardjohn: why bother having this if we already have CEL and its trivial to express in CEL?

howardjohn (changes requested):
I really think we should not do this.

1. I don't think "Kubernetes Agentic Networking" is an appropriate venue to define a general observability API. This has absolutely nothing to do with agentic networking specifically, and it feels like this is being used to "smuggle in" a standard through a low-barrier project so that it can gain momentum and face less scrutiny (just to be 100% clear: I am **not** accusing you of intentionally doing this, just that I think this is the likely result in practice).
2. Otel has proven two things, IMO: (1) making telemetry standards is really hard. (2) if you don't do it right, you won't make a standard but instead yet-another-non-standard (see all the pre-otel 'standards' that are now obsolete). (3) defining an API like this is hard/not something that can/should be done (hence why there is not only one).
3. I don't think this is really as portable as it seems on paper.

I'd like to explore a simple alternative: "implementations of agentic networking should implement the OTEL specification around traces, metrics, and logs. Implementations may provide ways to enable or disable these as needed".

I believe this will meet all of the use cases, without the daunting task of defining a new telemetry standard. From a user POV they just want these emitted from their agentic network; we do not need a new CRD to achieve that

david-martin:
> I really think we should not do this.
> 
> 1. I don't think "Kubernetes Agentic Networking" is an appropriate venue to define a general observability API. This has absolutely nothing to do with agentic networking specifically, and it feels like this is being used to "smuggle in" a standard through a low-barrier project so that it can gain momentum and face less scrutiny (just to be 100% clear: I am **not** accusing you of intentionally doing this, just that I think this is the likely result in practice).

Yeah, I can see where you are coming from. Others have raised similar concerns about there not being sufficient AI focused content (e.g. https://github.com/kubernetes-sigs/kube-agentic-networking/pull/69#pullrequestreview-3895841077).
I really don't think there's any ill intent or attempt to bypass/smuggle.

> 2. Otel has proven two things, IMO: (1) making telemetry standards is really hard. (2) if you don't do it right, you won't make a standard but instead yet-another-non-standard (see all the pre-otel 'standards' that are now obsolete). (3) defining an API like this is hard/not something that can/should be done (hence why there is not only one).

Agreed on 1.
Are there examples of 2 you could share so we can better understand why these have failed or are obsolute?
On 3, I'm still on the optimistic side that something could be done, but that being said, I'm not an implementer of a full gateway solution that would have to agree to and implement a standard API. 
Is the proposal too ambitious perhaps in its current form? Could it be slimmed down, but made extensible?
What if we started with just standardising 1 thing, like a spec for enabling/disabling the different signals and where to send them to? 

> 3. I don't think this is really as portable as it seems on paper.
> 
> I'd like to explore a simple alternative: "implementations of agentic networking should implement the OTEL specification around traces, metrics, and logs. Implementations may provide ways to enable or disable these as needed".

Do you see this being a central place to enable/disable or distributed?

gkhom:
> I don't think "Kubernetes Agentic Networking" is an appropriate venue to define a general observability API. This has absolutely nothing to do with agentic networking specifically, and it feels like this is being used to "smuggle in" a standard through a low-barrier project so that it can gain momentum and face less scrutiny (just to be 100% clear: I am not accusing you of intentionally doing this, just that I think this is the likely result in practice).

As this is my first ever contribution to the CNCF/K8s ecosystem, my intention is definitely not to bypass scrutiny or "smuggle" anything in. I opened this proposal here because I wanted to solve a real problem I've encountered (on several occasions) and this seemed like a reasonable starting point.

> if you don't do it right, you won't make a standard but instead yet-another-non-standard (see all the pre-otel 'standards' that are now obsolete).

I definitely want to avoid creating a fragmented or obsolete standard. Could you point me to some of the pre-OTel standardization attempts you mentioned? I'd love to understand where they fell short in defining a configuration API.

> defining an API like this is hard/not something that can/should be done

I recognize that standardizing this API carries significant complexity. Could you elaborate on the specific technical hurdles or roadblocks you envision? I'd really appreciate your perspective on what makes this intractable.

> I don't think this is really as portable as it seems on paper.

Could you provide some examples or edge cases where you see portability breaking down in this design? I'd like to see if we can address those gaps.

> I'd like to explore a simple alternative: "implementations of agentic networking should implement the OTEL specification around traces, metrics, and logs. Implementations may provide ways to enable or disable these as needed".

This would work as a baseline, but my concern is that it will fall short when users need to customize their telemetry (which is fairly common in practice). In absence of a standard, users will inevitably have to resort to vendor-specific APIs for any customization. Because a lot of operational machinery gets built around telemetry pipelines, relying on vendor-specific APIs for customization (re-)introduces the risk of vendor lock-in and becomes a blocker to portability. I think this is especially critical for agentic networking because it deals with autonomous tool execution. Telemetry here isn't just about troubleshooting and performance. It's a core requirement for operations, security, and auditability. If telemetry configuration is left entirely to vendor-specific implementations, it becomes nearly impossible for platform teams to guarantee a consistent audit trail across heterogeneous environments.

howardjohn (commented):
Looking at a lense of "what would a smaller but still powerful API look like":

targetRefs: duh, we need this. But do we need Namespace? Or just Mesh for the Mesh case? Or not worry about Mesh for now.

tracing:
* Do we need to support anything other than OTLP? OTLP won, its the de-facto and de-jure standard. Anything else is *probably* a waste of time, at least in the core API.
   * TracingProvider.endpoint should be a backendRef though
* Sampling is perhaps too simplistic -- will discuss more in a later comment as I need to run.
* context - do we need anything other than w3c? Its 2026.
* CustomAttribute: no doubt these are valuable but if we can only have literalValue its not. These only really provide value if we allow an ~expression language. Or a few wellknown attributes; I would say "just always include these" but for agentic use cases specifically this could include the entire prompt/completion which could warrant first class 'includeXYZ' style API. 

metrics: I think we just drop it entirely. TBH. Its pretty problematic to dynamically enable and disable metrics, especially on a per-metric or per-label-value basis. If we are talking prometheus there is also no config. If we need to support OTLP then just have `backendRef`.
* Definitely recommend dropping metrics override which is non-standard, hard to use, and hard to implement.

Logs:
* Notably missing otlp. Should it be? I know I said I was trying to reduce the scope :-) but it matters and influences other fields.
* Format -- not clear we need this. If its otlp, there is no format. If not, the proxy probably either uses structured logging and or does not, should a user really define that?
* match: nice to have but not strictly required. If we are going to do it, and are fine with CEL, just fully embrace CEL and don't have any other match types. If not, read https://blog.howardjohn.info/posts/cel-is-good/#cel-alternatives and think through whether a non-CEL based approach can make an acceptable user experience.
* fields: I don't think this can possibly just be a simple list of strings, and not a k/v pair, so hard to concretely discuss it but probably should be the same as CustomAttribute in tracing and the same concerns there apply.

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,317 @@
...
+type TelemetryPolicy struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    Spec TelemetryPolicySpec `json:"spec"`
+}
+
+type TelemetryPolicySpec struct {
+    // Identifies the target resources (Gateway or Namespace) to which this policy attaches (GEP-713).
+    TargetRefs []TargetReference `json:"targetRefs"`
```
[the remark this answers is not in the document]
gkhom: I'm leaning towards `NamespacedPolicyTargetReference` to allow cross-namespace configuration using `ReferenceGrants`. This would allow central management of uniform telemetry configuration.

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,340 @@
...
+
+  # 3. Access Logging
+  accessLogs:
+    enabled: true
+    matches: # Conditional logging
+      - cel: "response.code >= 500" # CEL-based filtering for errors
+    fields: # Configure specific fields to include
+      - "start_time"
+      - "response_code"
+      - "x-token-usage"
```
howardjohn: is there any definition of what these fields mean?

For a concrete example, say I want to log the MCP task name (I chose this since its not in https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/). Can I do it? what do I put as the field if I want to?

rikatz:
Can we move this discussion to Gateway API as a provisional GEP first, and then experimental? I think a lot of the discussion here is happening on a context that is important not only for agentic, but for the whole Gateway API ecosystem and I wouldn't like to receive this proposal as "we discussed and approved on agentic and now this needs to be implemented this way on Gateway API".

Thanks!

kflynn:
Seconding @rikatz's comment – this seems applicable to much more than just the agentic world, and I'd love to get eyes on it from Gateway API. Thanks!! 🙂

LiorLieberman:
@rikatz @kflynn have you had a chance to also review the API? 

There has been some really good iteration here. The fact that it probably belongs to Gateway is not questionable and have been discussed multiple times. 

Can you both review here, so we have the comments all cohesively in one place and we can move it to Gateway to the last round? Alternatively if any of you know of a way to move the proposal to with all the comments and history to Gateway then it will also be great. 

I would like to make sure Agentic usecases are represented, for this group this is the main motivation and its beyond "telemetry is useful to standardize generic cases as well" -- for agentic its an increasingly important feature and we need to make sure we iterate fast here and have the agentic cases are meaningfully represented.

rikatz:
No, I didn't mostly because I didn't knew it existed and heard about it recently on EGADS.

I will, but again, I will be strongly against this getting merged to Gateway API if this is merged to this repo first. I think it is a different audience, and the initial part of the GEP already says it is on the wrong place. Having all of these discussions here will make them be lost from the official GEP process.

```
A Kubernetes API for Gateway/Mesh Observability
```

docs/proposals/0069-TelemetryPolicy.md
aojea: Please do not use official apis without API approval, and as John says for examples use unambiguous domains

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,340 @@
...
+of how that traffic is observed.
+This deferral has led to "Observability Lock-in". Platform Engineering teams are forced to learn and manage distinct APIs for each environment. 
+A standardized `TelemetryPolicy` is necessary to decouple the intent of observability from the implementation. Without such standardization it is 
+difficult for platform owners to: 
+
+1. Enforce consistent auditing standards across different infrastructure providers.
+2. Support emerging workloads like AI Agents, which require specialized metrics (e.g., token usage, model latency) and detailed audit logs for tool-use verification.
+3. Manage “Mesh” and “Gateway” observability with a single unified API.
+
+## The Emergence of Agentic Networking
```
rikatz, line 28: do we really need a callout for this here, given that in the rest of the document is just specified that there is a need for a Telemetry API standardization, regardless of being Agentic Networking or not?

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,340 @@
...
+To address these challenges, the `TelemetryPolicy` proposal targets four core objectives:
+
+1. **Standardization**: A single API for Gateway and Mesh to configure Access Logging, Metrics generation, and Tracing propagation.
+2. **GEP-713 Compliance**: Support `targetRef` attachment to `Gateway` and `Namespace`. The latter covers Mesh use-cases.
+3. **Agentic Support**: Enable the capture of OpenTelemetry GenAI Semantic Conventions and support the requirements of PR #33.
+4. **Protocol Agnostic**: Support OpenTelemetry as the primary data model while allowing vendor-specific extensions.
+
+## The TelemetryPolicy Specification
+
+We propose the `TelemetryPolicy` as a direct policy attachment in the `agentic.networking.k8s.io` API group. See [GEP-713](https://gateway-api.sigs.k8s.io/geps/gep-713/#classes-of-policies) for more information on Direct attachment.
```
rikatz, line 52: I think this API is too generic to belong to this specific group, maybe start with x-k8s.io?

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,340 @@
...
+        type: Counter
+        dimensions: # Custom labels/dimensions
+          - key: "model_id"
+            fromHeader: "x-model-id" # Crucial for Agentic workloads
+
+  # 3. Access Logging
+  accessLogs:
+    enabled: true
+    matches: # Conditional logging
+      - cel: "response.code >= 500" # CEL-based filtering for errors
```
rikatz, line 100: One thing I am missing on this proposal is what is required and what is extended. Given this may become a Gateway API specification, we should at least define what is expected to be a core feature or not for this API.

eg.: The CEL matching here may not be implementable by all shippers.

docs/proposals/0069-TelemetryPolicy.md
rikatz, line 52: Also, WHY a policy attachment. Why are you chosing this approach instead of inline Gateway configuration? Can you be more explicit on your rationale behind this decision?

markfisher1:
No explicit requirement for:

* Policy decision trace (matched rule + context used)
* Delegation + tool invocation lineage
* Cross-agent correlation IDs

Recommend first-class “authorization trace graph” for debugging + compliance.

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,340 @@
...
+
+## The TelemetryPolicy Specification
+
+We propose the `TelemetryPolicy` as a direct policy attachment in the `agentic.networking.k8s.io` API group. See [GEP-713](https://gateway-api.sigs.k8s.io/geps/gep-713/#classes-of-policies) for more information on Direct attachment.
+
+### Resource Structure
+
+The following is an example that demonstrates the structure of the `TelemetryPolicy`.
+
+```yaml
```
rikatz, line 58: can we also have the status specification please?

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,340 @@
...
+      - attributeName: "env"
+        literalValue: "production"
+
+  # 2. Metrics Configuration
+  metrics:
+    enabled: true
+    overrides:
+      - name: "gateway.networking.k8s.io/http/request_count"
+        type: Counter
+        dimensions: # Custom labels/dimensions
```
rikatz, line 92: this part of the API is slightly confusing and seems somehow like a Prometheus rewrite rule. Is this something supported by OTEL, or some assumption that the Gateway implementation will do the rewrites/overrides here?

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,340 @@
...
+    Spec TelemetryPolicySpec `json:"spec"`
+
+    // status defines the observed state of TelemetryPolicy.
+    // +optional
+    Status TelemetryPolicyStatus `json:"status,omitempty"`
+}
+
+type TelemetryPolicySpec struct {
+    // Identifies the target resources (Gateway or Namespace) to which this policy attaches (GEP-713).
+    TargetRefs []NamespacedPolicyTargetReference `json:"targetRefs"`
```
rikatz, line 152: can I attach the same policy to both a Gateway and a namespace? what happens? Is there a precedence? How would the status look like in the case of a namespace attachment?

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,340 @@
...
+    // Configuration for access log generation.
+    AccessLogs *AccessLogsConfig `json:"accessLogs,omitempty"`
+}
+
+// --- Tracing Types ---
+
+type TracingConfig struct {
+
+    // Global switch to enable or disable tracing.
+    Enabled bool `json:"enabled"`
```
rikatz, line 169: Do not use bools on APIs. It is bad if you need something more than true/false (eg.: if you want a new Enabled semantic that means "partialEnable"). 

Please consider a different way to sate that this config is enabled or not (eg.: if tracingProvider is nil or not)

rikatz: (this comment applies to any usage of bool on this API)

docs/proposals/0069-TelemetryPolicy.md
```diff
@@ -0,0 +1,340 @@
...
+    // means the Policy is not relevant for any ancestors.
+    //
+    // If this slice is full, implementations MUST NOT add further entries.
+    // Instead they MUST consider the policy unimplementable and signal that
+    // on any related resources such as the ancestor that would be referenced
+    // here.
+    //
+    // +required
+    // +listType=atomic
+    // +kubebuilder:validation:MaxItems=16
```
rikatz, line 280: so the targetRef is also limited to 16?

youngnick:
Folks, I think it's very important to be clear. This is your subproject, and you can do what you want. But if you want to be able to take this upstream to Gateway API, you _must_ use the Gateway API GEP process, including the Provisional step.

We will _not_ be accepting a full API without full justification, using the _full_ process. So if you design and implement an API without using the Gateway API process, you're risking that this will never end up upstream. And none of us wants that, this is a _useful_ thing to have configured.

However, this reads like an attempt to force the Gateway API maintainers to do what you want by bring a completed API and implementations to us, and then saying "why are you being so difficult". Maybe it is, maybe it isn't, but that's what it looks like.

We have _very_ good reasons to be wary of adding Policy objects to Gateway API. They are _very_ difficult to get right, and I have huge concerns with using the same Policy object for two _very different_ targets. This Policy is also _clearly_ useful to other usecases aside from Agentic Networking, and so _must_ be designed for those use cases as well.

LiorLieberman:
The conversation is getting heated without a reason. @youngnick whats the concrete concern? That the template is not Gateway template verbatim? 

There are good comments from Ricardo about some missing sections - like why policy and a few others. And we are in agreement that we move this proposal to Gateway (was also discussed in EGADS iiuc).

I am ooo so was slow to respond on gh but chatted with Ricardo on slack this morning and confirmed that as well.

Regarding different layers @youngnick, I agree with you. It does look like starting with Gateway is more simple and more useful, at least for the time being. But happy to hear your thoughts as well. 

Agentic Net has concrete needs (and it will continue to have such), a lot of the work here will intersect with Gateway, some will apis will intersect more and some will intersect less. It's ok to have the "is Gateway the right home for this api" discussions and its part of the process.

Anyways, i promised to move that to Gateway EOW.

youngnick:
I've given this a better read, and here's my thoughts, more specifically:

tl;dr This Policy, as written, is not suitable for upstreaming into Gateway API as it stands.

It solves a problem that exists for most if not all Gateway API users, but builds the agentic parts into the core API. I'd much, much rather see a solution that allows for configuration of a generic telemetry config across HTTP requests (that is, requests where the Gateway implementation has access to an unencrypted HTTP stream), with an extension point (that can be pre-filled with agentic extensions, sure).

I'd also question, again, why it needs to be a Policy. The document assumes that the only way to do this is with Policy, which is true when the code lives in this repo, under this subproject, but is not the case upstream in Gateway API. If we want to configure all the HTTP traffic passing through a Gateway, why not add a top-level struct on the Gateway to do that? We've already done that recently for TLS, and telemetry is a core enough requirement that it makes sense. Maybe it's better to have the sink parameters and other definitely-global config at the Gateway level, and extend the fiddlier bits with something else.

But, that conversation is not really possible when we're not having it in the Gateway API process. So, I was a little forceful before, but I am telling you all that what will definitely not happen is building something here, and then bringing that thing upstream to Gateway and expecting it to get merged unchanged.

I also agree that if you want to target Mesh use cases, then the Mesh resource is probably the right thing to target - but again, that is a thing where a Policy is not required, because once the proposal is in upstream Gateway API, we can inline the fields inside the Mesh resource.

I would also strongly recommend that if you must have a Policy, and are going to allow this Policy to attach to multiple kinds of target, you carefully consider interaction and conflict management. What happens when there are two Policies targeting the same thing? What happens when a Gateway is in a namespace that is managed by a mesh? Do both configs take effect? If not, which one wins? @guicassolato mentions a merge strategy of `None`, which implies the latter, and which also means that you should use first-object-created conflict resolution, as upstream Gateway API does.

I also looked over @rikatz's API review, and agree with his comments.

In particular, `bool` in Kubernetes APIs is a bad idea, not only because of the concern he lists about extensibility, but also because of the annoying handling of truth-y values in YAML. It's _way_ better to use a `mode` enum than a `modeEnabled` bool, in almost all cases.

kflynn:
I have also reread and broadly agree with @youngnick and @rikatz. I'm not going to say much about specific details, because to me those are wildly overwhelmed by the fact that GEP-713 policy is very much not my first choice for how to approach this -- and in my mind, that immediately brings us back around to questions about upstream Gateway API and its processes.

Many of y'all participate in Gateway API pretty actively, but for the benefit of those who don't: if you want to add something to Gateway API - and my sense is that many or most of y'all would like this to be part of Gateway API itself - you start with a Provisional GEP. Provisional GEPs aren't about the details of the API, they're about the big questions: what problems are we trying to solve? why are they important to solve? for whom are we designing the solutions? These are critical, and they're rooted in explorations of user stories rather than API design -- but questions like "should we use GEP-713 policy for this?" are wrapped up in those questions, too.

So the first thing I'm saying here is that given that feeling that there seems to be consensus around moving into Gateway API, a Provisional GEP feels like a better next step than anything else -- because the _other_ thing is that without the Provisional GEP, there isn't a way into Gateway API, and if someone were to present an API design, we would need to back up to the Provisional step where everything is back on the table _anyway_, because though many of y'all may already have the answers to what, why, and who in your head, the Gateway API folks don't, and will need it explained.

Additionally, as for _why_ GEP-713 isn't my first choice... GEP-713 is really a way of coping with the fact that we can't extend core Kubernetes resources like Service. When we're dealing with resources we _do_ control - like Gateways or Routes - GEP-713 should never be the first thing to reach for. Policies are deceptively easy to create, but they introduce really nasty questions (as @youngnick already mentioned) and every last one we accept as a part of Gateway API itself comes with the burden of supporting it - with all its issues - forever. 🙁 That's why you're seeing pushback there from me, and from others.

david-martin:
For avoidance of doubt, a GEP is/will be opened in gateway-api for this.

/remove-lgtm
/remove-approve
/hold

Keeping open for any remaining agentic specific discussion to conclude before closing.

sjberman:
Wanted to chime in since I don't see NGINX considered yet. We currently support OTEL tracing through an [ObservabilityPolicy](https://docs.nginx.com/nginx-gateway-fabric/reference/api/#gateway.nginx.org/v1alpha2.ObservabilityPolicy). This enables tracing for a Route.

We also have a global nginx CRD where a user would set their exporter URL for the Gateway. This design was originally so a user didn't have to set this value on every single policy, and instead just set it globally. This CRD also allows for setting global span attributes for the Gateway.

[Here](https://docs.nginx.com/nginx-gateway-fabric/monitoring/tracing/) is our tracing document on how we set everything up.

We also have considerations for [future enhancements](https://github.com/nginx/nginx-gateway-fabric/issues/3276) to our own module.

david-martin:
@gkhom OK to close this now?
I see https://github.com/kubernetes-sigs/gateway-api/pull/4872 is open upstream in gwapi

LiorLieberman:
closing in favor of https://github.com/kubernetes-sigs/gateway-api/pull/4872
/close

PR #76, opened 2026-02-10
Use latest main tag for images

chuangw6:
**What type of PR is this?**

/kind bug

Using latest main tag for images helps avoid updating the image tag every time we update the code as we are actively developing the controller and agent components.

We can switch to a specific workable tag once we are ready to demo it.

```release-note
None
```


LiorLieberman:
since its just temporary, sgtm. 

/approve
/lgtm
/ok-to-test

PR #133, opened 2026-03-02
doc: add readme to document agent identify

chuangw6:
**What type of PR is this?**

/kind documentation

add a readme to explicitly document how users can configure their agents for mTLS using KAN's native Kubernetes identity features.

```release-note
NONE
```

### DIFF
--- a/site-src/guides/quickstart/agent-identity-demo/README.md
+++ b/site-src/guides/quickstart/agent-identity-demo/README.md
@@ -1,0 +1,75 @@
+# Agent Identity Demo
+
+This demo illustrates how to configure a Kubernetes workload (AI Agent) to receive an mTLS identity certificate signed by the Kube Agentic Networking (KAN) identity signer.
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


quickstart/agent-identity-demo/README.md
```diff
@@ -0,0 +1,71 @@
...
+    kubectl apply -f quickstart/agent-identity-demo/agent-identity-demo.yaml
+    ```
+3.  Wait for the client pod to be ready:
+    ```bash
+    kubectl wait --for=condition=Ready pod -l app=client -n agent-identity-demo
+    ```
+4.  Verify the certificates are mounted:
+    ```bash
+    kubectl exec -it deployment/client -n agent-identity-demo -- ls /run/agent-identity-mtls
+    ```
```
haiyanmeng: Can we add another command for checking the SPIFFE ID? It should be sth like (please verify the command, thanks!):
```
kubectl exec $POD -n $NS -c $CONTAINER  -- \
  cat /run/agent-identity-mtls/credential-bundle.pem | \
  openssl x509 --noout --text | \
  grep "URI:"
```

Issue #75, opened 2026-02-10
Improve tool calling with LLM unpredictability


chuangw6:
In the current quickstart, which uses the DeepSeek model from Hugging Face, the agent occasionally processes a prompt (e.g., "Can you do 1 + 1?") internally instead of invoking the available MCP tools. 

This issue has never occurred with the Google gemini-2.5-flash model. 

We should improve this behavior to avoid confusion - we likely need to find a better model from HF that supports both chat and tool calling.

ziyue-101:
We might be able to force agents to trigger MCP tools no matter what. Some sdks allow developers attaching a set of tools into users prompt message. I will update the examples.

/assign @ziyue-101

PR #137, opened 2026-03-03
improve the system prompt to reduce model halluciation

ziyue-101:
/kind feature

Ref #75

```release-note
None
```


haiyanmeng:
> improve the system promot to reduce model halluciation

@ziyue-101 , there is typo in the PR title. `promot` should be `prompt`.

PR #177, opened 2026-03-13
Fix default build target and some minor issues in the Makefile

bowei:
Fix default build target and some minor issues in the Makefile
- First target in the Makefile is the default target
- Remove references to uninitialized variables

/kind cleanup


LiorLieberman (approved):
Sweet, thanks @bowei !
/lgtm

PR #198, opened 2026-03-20
Fix Gateway ObservedGeneration timeouts and support infrastructure labels/annotations

haiyanmeng:
**What type of PR is this?**

/kind bug

This commit addresses the following issues:
    
    1. Fixes an issue where the syncGateway reconciliation loop was bypassed
       during proxy provisioning (EnsureProxyExist) or xDS translation
       failures. This ensures the Gateway's ObservedGeneration correctly
       updates its status to Programmed: False instead of skipping the
       update entirely.
    
    2. Implements propagation of Gateway Spec.Infrastructure Labels and
       Annotations down to the generated Envoy proxy Deployments, Pods,
       Services, and ServiceAccounts.
    
    3. Refactors error handling in syncGateway to collect and merge errors
       from status updates and use standard rate-limiting backoff on proxy
       creation failures.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/constants/envoynames.go
+++ b/pkg/constants/envoynames.go
@@ -38,7 +38,8 @@ const (
 	EnvoyBootstrapMountPath = "/etc/envoy/bootstrap"
 
 	// GatewayNameLabel is the label key used to identify resources associated with a specific Gateway.
-	GatewayNameLabel = ProjectName + "/gateway-name"
+	// This label was formally introduced in https://gateway-api.sigs.k8s.io/geps/gep-1762/#automated-deployments.
+	GatewayNameLabel = "gateway.networking.k8s.io/gateway-name"
 
 	// GatewayRBACFilterNamePrefix is the prefix for Envoy RBAC filters at the gateway level.
 	GatewayRBACFilterNamePrefix = "envoy.filters.http.rbac.gateway_level_"
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -400,7 +400,8 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
 	rm := envoy.NewResourceManager(c.core.client, gateway, c.envoyImage, c.agenticIdentityTrustDomain)
 	proxyIP, err := rm.EnsureProxyExist(ctx)
 	if err != nil {
-		return err
+		updateErr := updateGatewayStatus(ctx, c, gateway, newGW, nil, err)
+		return errors.Join(err, updateErr)
 	}
 
 	// TODO: Add support for IPv6?
@@ -418,25 +419,18 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
 	// Translate Gateway to xDS resources (includes only current HTTPRoutes/XAccessPolicies; stale rules are omitted).
 	resources, listenerStatuses, httpRouteStatuses, _, err := c.translator.TranslateGatewayToXDS(ctx, gateway)
 	if err != nil {
-		// TODO: Update Gateway status with the error.
-		return fmt.Errorf("failed to translate gateway to xDS resources: %w", err)
+		updateErr := updateGatewayStatus(ctx, c, gateway, newGW, nil, fmt.Errorf("failed to translate gateway to xDS resources: %w", err))
+		return errors.Join(err, updateErr)
 	}
 
 	logger.Info("Translated gateway to xDS resources")
 
 	newGW.Status.Listeners = listenerStatuses
 	// Update the xDS server with the new resources.
-	err = c.xdsServer.UpdateXDSServer(ctx, rm.NodeID(), resources)
-
-	setGatewayConditions(newGW, listenerStatuses, err)
-
-	if !reflect.DeepEqual(gateway.Status, newGW.Status) {
-		if err := c.updateGatewayStatus(ctx, newGW.Namespace, newGW.Name, &newGW.Status); err != nil {
-			return fmt.Errorf("failed to update gateway status: %w", err)
-		}
-		logger.Info("Updated gateway status")
-	}
-	return c.updateHTTPRouteStatuses(ctx, httpRouteStatuses)
+	xdsErr := c.xdsServer.UpdateXDSServer(ctx, rm.NodeID(), resources)
+	updateGatewayStatusErr := updateGatewayStatus(ctx, c, gateway, newGW, listenerStatuses, xdsErr)
+	updateHTTPRouteStatusErr := c.updateHTTPRouteStatuses(ctx, httpRouteStatuses)
+	return errors.Join(xdsErr, updateGatewayStatusErr, updateHTTPRouteStatusErr)
 }
 
 func (c *Controller) updateGatewayRemoveFinalizer(ctx context.Context, namespace, name string) error {
@@ -455,6 +449,23 @@ func (c *Controller) updateGatewayRemoveFinalizer(ctx context.Context, namespace
 		_, err = c.gateway.client.GatewayV1().Gateways(namespace).Update(ctx, u, metav1.UpdateOptions{})
 		return err
 	})
+}
+
+// updateGatewayStatus updates the Gateway status in the API server if it has changed.
+// Note: this function modifies newGW.
+func updateGatewayStatus(ctx context.Context, c *Controller, gateway *gatewayv1.Gateway, newGW *gatewayv1.Gateway, listenerStatuses []gatewayv1.ListenerStatus, syncErr error) error {
+	setGatewayConditions(newGW, listenerStatuses, syncErr)
+
+	if reflect.DeepEqual(gateway.Status, newGW.Status) {
+		return nil
+	}
+
+	if _, statusErr := c.gateway.client.GatewayV1().Gateways(newGW.Namespace).UpdateStatus(ctx, newGW, metav1.UpdateOptions{}); statusErr != nil {
+		klog.FromContext(ctx).Error(statusErr, "failed to update gateway status")
+		return fmt.Errorf("failed to update gateway status: %w", statusErr)
+	}
+	klog.FromContext(ctx).Info("Updated gateway status")
+	return nil
 }
 
 // ensureGatewayFinalizer adds the controller finalizer via the API when missing.
@@ -490,22 +501,6 @@ func (c *Controller) ensureGatewayFinalizer(ctx context.Context, namespace, name
 	}
 	nowHas := sets.New(fresh.Finalizers...).Has(constants.GatewayFinalizer)
 	return !hadFinalizer && nowHas, nil
-}
-
-func (c *Controller) updateGatewayStatus(ctx context.Context, namespace, name string, desired *gatewayv1.GatewayStatus) error {
-	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
-		latest, err := c.gateway.gatewayLister.Gateways(namespace).Get(name)
-		if err != nil {
-			return err
-		}
-		if reflect.DeepEqual(latest.Status, *desired) {
-			return nil
-		}
-		u := latest.DeepCopy()
-		u.Status = *desired.DeepCopy()
-		_, err = c.gateway.client.GatewayV1().Gateways(namespace).UpdateStatus(ctx, u, metav1.UpdateOptions{})
-		return err
-	})
 }
 
 // hasHTTPRoutesReferencingGateway returns true if any HTTPRoute has a ParentRef to the given Gateway.
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2025 The Kubernetes Authors.
+Copyright The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -17,16 +17,25 @@ limitations under the License.
 package controller
 
 import (
+	"context"
+	"fmt"
 	"testing"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
 
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewayfake "sigs.k8s.io/gateway-api/pkg/client/clientset/versioned/fake"
 	gatewaylisters "sigs.k8s.io/gateway-api/pkg/client/listers/apis/v1"
 
 	agenticv0alpha0 "sigs.k8s.io/kube-agentic-networking/api/v0alpha0"
 	agenticlisters "sigs.k8s.io/kube-agentic-networking/k8s/client/listers/api/v0alpha0"
+	"sigs.k8s.io/kube-agentic-networking/pkg/constants"
 )
 
 func TestHasHTTPRoutesReferencingGateway(t *testing.T) {
@@ -341,4 +350,92 @@ func ptrGatewayNamespace(ns string) *gatewayv1.Namespace {
 func ptrGatewayKind(kind string) *gatewayv1.Kind {
 	k := gatewayv1.Kind(kind)
 	return &k
+}
+
+type fakeQueue struct {
+	workqueue.TypedRateLimitingInterface[string]
+	addedAfter string
+	delay      time.Duration
+}
+
+func (q *fakeQueue) AddAfter(item string, duration time.Duration) {
+	q.addedAfter = item
+	q.delay = duration
+}
+
+func (q *fakeQueue) Len() int {
+	if q.addedAfter != "" {
+		return 1
+	}
+	return 0
+}
+
+func TestSyncGateway_EnsureProxyExistError(t *testing.T) {
+	gwNamespace := "default"
+	gwName := "my-gateway"
+	gwcName := "my-class"
+
+	gw := &gatewayv1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{Namespace: gwNamespace, Name: gwName, Finalizers: []string{constants.GatewayFinalizer}},
+		Spec: gatewayv1.GatewaySpec{
+			GatewayClassName: gatewayv1.ObjectName(gwcName),
+		},
+	}
+
+	gwc := &gatewayv1.GatewayClass{
+		ObjectMeta: metav1.ObjectMeta{Name: gwcName},
+		Spec: gatewayv1.GatewayClassSpec{
+			ControllerName: constants.ControllerName,
+		},
+	}
+
+	gwIndexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+	if err := gwIndexer.Add(gw); err != nil {
+		t.Fatalf("gwIndexer.Add() = %v, want nil", err)
+	}
+
+	gwcIndexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
+	if err := gwcIndexer.Add(gwc); err != nil {
+		t.Fatalf("gwcIndexer.Add() = %v, want nil", err)
+	}
+
+	fakeK8sClient := fake.NewClientset()
+	// Force error on SA creation
+	fakeK8sClient.PrependReactor("create", "serviceaccounts", func(_ k8stesting.Action) (handled bool, ret runtime.Object, err error) {
+		return true, nil, fmt.Errorf("forced error")
+	})
+
+	fakeGwClient := gatewayfake.NewClientset(gw)
+
+	queue := &fakeQueue{}
+
+	c := &Controller{
+		gateway: gatewayResources{
+			gatewayLister:      gatewaylisters.NewGatewayLister(gwIndexer),
+			gatewayClassLister: gatewaylisters.NewGatewayClassLister(gwcIndexer),
+			client:             fakeGwClient,
+		},
+		core: coreResources{
+			client: fakeK8sClient,
+		},
+		gatewayqueue: queue,
+	}
+
+	err := c.syncGateway(context.Background(), gwNamespace+"/"+gwName)
+	if err == nil {
+		t.Fatal("syncGateway() = nil, want error")
+	}
+
+	// Verify status was updated with error!
+	actions := fakeGwClient.Actions()
+	foundUpdateStatus := false
+	for _, action := range actions {
+		if action.GetVerb() == "update" && action.GetSubresource() == "status" {
+			foundUpdateStatus = true
+			break
+		}
+	}
+	if !foundUpdateStatus {
+		t.Error("foundUpdateStatus = false, want true")
+	}
 }
--- a/pkg/infra/envoy/proxy.go
+++ b/pkg/infra/envoy/proxy.go
@@ -21,14 +21,12 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
-	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 
@@ -95,6 +93,7 @@ func (r *ResourceManager) NodeID() string {
 	return r.nodeID
 }
 
+// ensureSA ensures that the ServiceAccount for the Envoy proxy exists.
 func (r *ResourceManager) ensureSA(ctx context.Context) error {
 	logger := klog.FromContext(ctx)
 
@@ -102,6 +101,7 @@ func (r *ResourceManager) ensureSA(ctx context.Context) error {
 	_, err := r.client.CoreV1().ServiceAccounts(sa.Namespace).Get(ctx, sa.Name, metav1.GetOptions{})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
+			logger.Info("Creating Envoy proxy serviceaccount", "name", sa.Name, "namespace", sa.Namespace)
 			_, err = r.client.CoreV1().ServiceAccounts(sa.Namespace).Create(ctx, sa, metav1.CreateOptions{})
 			if err != nil {
 				return fmt.Errorf("failed to create envoy serviceaccount: %w", err)
@@ -114,6 +114,7 @@ func (r *ResourceManager) ensureSA(ctx context.Context) error {
 	return nil
 }
 
+// ensureConfigMap ensures that the ConfigMap for the Envoy proxy exists.
 func (r *ResourceManager) ensureConfigMap(ctx context.Context) error {
 	logger := klog.FromContext(ctx)
 	cm, err := r.renderConfigMap()
@@ -124,6 +125,7 @@ func (r *ResourceManager) ensureConfigMap(ctx context.Context) error {
 	_, err = r.client.CoreV1().ConfigMaps(cm.Namespace).Get(ctx, cm.Name, metav1.GetOptions{})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
+			logger.Info("Creating Envoy bootstrap configmap", "name", cm.Name, "namespace", cm.Namespace)
 			_, err = r.client.CoreV1().ConfigMaps(cm.Namespace).Create(ctx, cm, metav1.CreateOptions{})
 			if err != nil {
 				return fmt.Errorf("failed to create envoy configmap: %w", err)
@@ -137,14 +139,16 @@ func (r *ResourceManager) ensureConfigMap(ctx context.Context) error {
 	return nil
 }
 
+// ensureDeployment ensures that the Envoy deployment exists and is available.
 func (r *ResourceManager) ensureDeployment(ctx context.Context) error {
 	logger := klog.FromContext(ctx)
 
 	deployment := r.renderDeployment()
-	_, err := r.client.AppsV1().Deployments(deployment.Namespace).Get(ctx, deployment.Name, metav1.GetOptions{})
+	dep, err := r.client.AppsV1().Deployments(deployment.Namespace).Get(ctx, deployment.Name, metav1.GetOptions{})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			_, err = r.client.AppsV1().Deployments(deployment.Namespace).Create(ctx, deployment, metav1.CreateOptions{})
+			logger.Info("Creating Envoy proxy deployment", "name", deployment.Name, "namespace", deployment.Namespace)
+			dep, err = r.client.AppsV1().Deployments(deployment.Namespace).Create(ctx, deployment, metav1.CreateOptions{})
 			if err != nil {
 				return fmt.Errorf("failed to create envoy deployment: %w", err)
 			}
@@ -153,19 +157,26 @@ func (r *ResourceManager) ensureDeployment(ctx context.Context) error {
 		}
 	}
 
-	if err := waitForDeploymentAvailable(ctx, r.client, deployment.Namespace, deployment.Name); err != nil {
-		return err
+	// If the Deployment was just created, we will highly likely immediately fail
+	// here and return.
+	for _, cond := range dep.Status.Conditions {
+		if cond.Type == appsv1.DeploymentAvailable && cond.Status == corev1.ConditionTrue {
+			logger.Info("Envoy proxy deployment is ready!")
+			return nil
+		}
 	}
-	logger.Info("Envoy proxy deployment is ready!")
-	return nil
+
+	return fmt.Errorf("envoy deployment %s is not available yet", deployment.Name)
 }
 
+// ensureService ensures that the Service for the Envoy proxy exists and has a ClusterIP assigned.
 func (r *ResourceManager) ensureService(ctx context.Context) (string, error) {
 	logger := klog.FromContext(ctx)
 	service := r.renderService()
 	svc, err := r.client.CoreV1().Services(service.Namespace).Get(ctx, service.Name, metav1.GetOptions{})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
+			logger.Info("Creating Envoy proxy service", "name", service.Name, "namespace", service.Namespace)
 			svc, err = r.client.CoreV1().Services(service.Namespace).Create(ctx, service, metav1.CreateOptions{})
 			if err != nil {
 				return "", fmt.Errorf("failed to create envoy service: %w", err)
@@ -175,12 +186,6 @@ func (r *ResourceManager) ensureService(ctx context.Context) (string, error) {
 		}
 	}
 
-	if errWait := waitForServiceReady(ctx, r.client, service.Namespace, service.Name); errWait != nil {
-		return "", errWait
-	}
-	logger.Info("Envoy proxy service is ready!")
-
-	// Refresh the service object to get the assigned ClusterIP if it was just created.
 	if svc.Spec.ClusterIP == "" {
 		svc, err = r.client.CoreV1().Services(service.Namespace).Get(ctx, service.Name, metav1.GetOptions{})
 		if err != nil {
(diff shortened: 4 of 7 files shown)


pkg/controller/controller.go
```diff
@@ -379,11 +379,28 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
...
+		return syncErr
+	}
+
 	// Ensure Envoy proxy deployment and service exist.
 	rm := envoy.NewResourceManager(c.core.client, gateway, c.envoyImage, c.agenticIdentityTrustDomain)
 	proxyIP, err := rm.EnsureProxyExist(ctx)
 	if err != nil {
-		return err
+		_ = updateStatusFunc(nil, err)
+		c.gatewayqueue.AddAfter(key, 2*time.Second)
```
LiorLieberman: can you explain the rationale for adding 2 seconds?

I can read the below from the PR description, but not sure I really follow. Were there a ctx deadline exceeded errors? And how the 2 sec help here?

>Combined with a fixed 2-second requeue in the controller, this prevents worker thread exhaustion and Context Deadline Exceeded errors on the client-go API client

LiorLieberman (commented):
This is great, thanks @haiyanmeng. 
I left some questions, and would appreciate adding unitests for the logic

pkg/controller/controller.go
```diff
@@ -379,11 +379,28 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
 
 	logger.Info("Syncing gateway")
 
+	updateStatusFunc := func(listenerStatuses []gatewayv1.ListenerStatus, syncErr error) error {
```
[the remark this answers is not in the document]
haiyanmeng: `updateStatusFunc` acts as a commit function for all the status changes made to `newGW.Status` throughout the `syncGateway` flow.

pkg/controller/controller.go
haiyanmeng: Because `EnsureProxyExist` fails fast when the Envoy deployment or service is still spinning up, returning that error to the controller would normally cause the item to be re-enqueued immediately (depending on the rate limiter). This creates a tight loop that spams the API server with Get requests. By using `AddAfter(key, 2*time.Second)`, we explicitly tell the controller to wait 2 seconds before trying again, giving the Kubernetes control plane and nodes time to make progress on creating the resources.

pkg/infra/envoy/proxy.go
```diff
@@ -192,43 +190,27 @@ func (r *ResourceManager) ensureService(ctx context.Context) (string, error) {
 }
 
 func waitForServiceReady(ctx context.Context, client kubernetes.Interface, namespace, name string) error {
```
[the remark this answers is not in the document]
haiyanmeng: The retry logic was replaced by this asynchronous workqueue-based retry mechanism. This is the standard pattern for Kubernetes controllers as it avoids blocking the controller's worker threads while waiting for external resources to become ready.

pkg/infra/envoy/resourcerender.go
```diff
@@ -108,6 +108,31 @@ func generateSdsConfig(tmpl, sdsConfigName, trustDomain string) (string, error)
...
+func (r *ResourceManager) getLabels() map[string]string {
+	labels := map[string]string{
+		constants.GatewayNameLabel: r.gw.Name,
+	}
+	if r.gw.Spec.Infrastructure != nil {
+		for k, v := range r.gw.Spec.Infrastructure.Labels {
+			labels[string(k)] = string(v)
+		}
+	}
+	return labels
```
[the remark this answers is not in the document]
haiyanmeng: If you look at `ensureDeployment` in pkg/infra/envoy/proxy.go (and similarly for Service and ConfigMap), you'll notice that we only create the resource if it does not exist. We do not currently have any logic to update the resource if it already exists:

Because we never call Update on the Deployment, we will not overwrite any labels added by users or other controllers on an existing Deployment.

pkg/infra/envoy/proxy.go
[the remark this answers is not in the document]
haiyanmeng: I refactored ensureService and removed waitForServiceReady; and refactored ensureDeployment and removed waitForDeploymentAvailable.

haiyanmeng:
> would appreciate adding unitests for the logic

Done

pkg/infra/envoy/resourcerender.go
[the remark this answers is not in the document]
LiorLieberman: oh really, you mean that we create it and not reconciling it further? I.E if someone changes the image of the deployment, we would not do anything about it?

If yes, do you know why did we choose to do it? I see this deployment as a "managed" deployment, its is created from Gateway - and the user, generally should not deal with it, with the exception of labels, which may be common for some users.

pkg/controller/controller.go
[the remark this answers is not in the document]
LiorLieberman: I understand, but this func only receives ListenerStatus array. So where are we updating other components on Gateway Stauts?

pkg/controller/controller.go
LiorLieberman: Thanks. Did we see CTX deadline exceeded errors in the controller logs?

pkg/controller/controller.go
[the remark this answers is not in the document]
haiyanmeng: `updateStatusFunc` is an inline function (more precisely, an anonymous function or closure) defined within the `syncGateway` method. It will pick up `newGW`. In Go, closures capture variables by reference. Since `updateStatusFunc` references newGW, it captures the `newGW` variable itself.

pkg/controller/controller.go
haiyanmeng: The reason for this was the conformance test failure. I didn't check the controller logs before making this change.

pkg/infra/envoy/resourcerender.go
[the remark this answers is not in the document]
haiyanmeng: I believe this code was written in this way for simplicity. 

I created https://github.com/kubernetes-sigs/kube-agentic-networking/issues/238 to track the gap.

pkg/controller/controller.go
LiorLieberman: How do we know this is the fix to the issue then?

haiyanmeng: The conformance test passed and the e2e test passed with this change.

pkg/controller/controller.go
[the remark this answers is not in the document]
haiyanmeng: Refactored the updateStatusFunc closure in pkg/controller/controller.go into a standalone function named updateGatewayStatus

pkg/constants/envoynames.go
```diff
@@ -38,7 +38,7 @@ const (
 	EnvoyBootstrapMountPath = "/etc/envoy/bootstrap"
 
 	// GatewayNameLabel is the label key used to identify resources associated with a specific Gateway.
-	GatewayNameLabel = ProjectName + "/gateway-name"
+	GatewayNameLabel = "gateway.networking.k8s.io/gateway-name"
```
bowei, line 42: Is this part of the GW API as a standardized label? It would be good to include a reference to the doc (I wasn't able to find it).

pkg/infra/envoy/proxy.go
```diff
@@ -192,43 +190,27 @@ func (r *ResourceManager) ensureService(ctx context.Context) (string, error) {
...
-		svc, err := client.CoreV1().Services(namespace).Get(ctx, name, metav1.GetOptions{})
-		if err != nil {
-			return false, err
-		}
-		if svc.Spec.ClusterIP != "" {
-			return true, nil
-		}
-		return false, nil
-	})
+	svc, err := client.CoreV1().Services(namespace).Get(ctx, name, metav1.GetOptions{})
```
bowei: This replaces a poll wait with just returning? Seems wrong. You will need to change the name of the function if this was the intent.

pkg/infra/envoy/proxy.go
```diff
@@ -192,43 +190,27 @@ func (r *ResourceManager) ensureService(ctx context.Context) (string, error) {
...
+		return err
 	}
-	return nil
+	if svc.Spec.ClusterIP != "" {
+		return nil
+	}
+	return fmt.Errorf("envoy service %s is not ready yet", name)
 }
 
 func waitForDeploymentAvailable(ctx context.Context, client kubernetes.Interface, namespace, name string) error {
```
bowei: Same here. We can't just change the wait func to not wait. Either change the function name or this fix is incorrect.

pkg/infra/envoy/resourcerender.go
```diff
@@ -108,6 +108,31 @@ func generateSdsConfig(tmpl, sdsConfigName, trustDomain string) (string, error)
...
+}
+
+func (r *ResourceManager) getAnnotations() map[string]string {
```
bowei: This should be a helper not attached to ResourceManager taking in Infrastructure.Annotations as the argument to the function.

pkg/infra/envoy/resourcerender.go
```diff
@@ -108,6 +108,31 @@ func generateSdsConfig(tmpl, sdsConfigName, trustDomain string) (string, error)
 	return buff.String(), nil
 }
 
+func (r *ResourceManager) getLabels() map[string]string {
```
bowei: This should be a free helper method that takes the Labels as an argument.

bowei: This func should have associated unit tests.

pkg/infra/envoy/resourcerender.go
bowei: This func should have associated unit tests.

pkg/infra/envoy/proxy.go
```diff
@@ -151,11 +151,14 @@ func (r *ResourceManager) ensureDeployment(ctx context.Context) error {
 		}
 	}
 
-	if err := waitForDeploymentAvailable(ctx, r.client, deployment.Namespace, deployment.Name); err != nil {
-		return err
+	for _, cond := range dep.Status.Conditions {
```
bowei, line 162: ? don't we need to wait?

bowei (commented):
I don't understand how this change works. Maybe we can walk through it live?

pkg/constants/envoynames.go
haiyanmeng: Yes. It was formally introduced as part of GEP-1762 (Gateway Enhancement Proposal) to standardize how Gateway controllers label the underlying infrastructure they manage. https://gateway-api.sigs.k8s.io/geps/gep-1762/#automated-deployments

haiyanmeng: Let me add the GEP link as a comment

pkg/infra/envoy/resourcerender.go
haiyanmeng: Done

pkg/infra/envoy/resourcerender.go
haiyanmeng: Done

pkg/infra/envoy/proxy.go
haiyanmeng: Before this PR, there was 1-minute wait.PollUntilContextTimeout loop in waitForServiceReady and waitForDeploymentAvailable, and the wait was blocking.

This PR replaces the blocking wait with a fast, non-blocking check. Combined with a fixed 2-second requeue in the controller, this prevents worker thread exhaustion and Context Deadline Exceeded errors on the client-go API client.

pkg/infra/envoy/proxy.go
haiyanmeng: It seems you commented on an outdated version of the code. The function was named to `ensureDeployment`. I added a comment to clarify that these functions do not implement polling or waiting loops for resource availability.  Instead, they return an error if a resource is not ready, relying on the gateway controller queue to retry reconciliation after a delay.

pkg/infra/envoy/proxy.go
haiyanmeng: It seems you commented on an outdated version of the code. The function was named to `ensureService`. I added a comment to clarify that these functions do not implement polling or waiting loops for resource availability.  Instead, they return an error if a resource is not ready, relying on the gateway controller queue to retry reconciliation after a delay.

pkg/controller/controller.go
```diff
@@ -400,7 +400,9 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
 	rm := envoy.NewResourceManager(c.core.client, gateway, c.envoyImage, c.agenticIdentityTrustDomain)
 	proxyIP, err := rm.EnsureProxyExist(ctx)
 	if err != nil {
-		return err
+		_ = updateGatewayStatus(ctx, c, gateway, newGW, nil, err)
+		c.gatewayqueue.AddAfter(key, 2*time.Second)
```
bowei: Why isn't this just return an error and let it backoff?

I also think we should just have the controller startup wait for the Envoy to be deployed? Wouldn't that take care of a lot of initialization ordering issues?

pkg/controller/controller.go
```diff
@@ -418,23 +420,19 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
...
-	setGatewayConditions(newGW, listenerStatuses, err)
-
-	if !reflect.DeepEqual(gateway.Status, newGW.Status) {
-		if err := c.updateGatewayStatus(ctx, newGW.Namespace, newGW.Name, &newGW.Status); err != nil {
-			return fmt.Errorf("failed to update gateway status: %w", err)
-		}
-		logger.Info("Updated gateway status")
+	if err := updateGatewayStatus(ctx, c, gateway, newGW, listenerStatuses, xdsErr); err != nil {
+		// Attempt to update route statuses anyway, as translation was done, but return the xds/status err
+		_ = c.updateHTTPRouteStatuses(ctx, httpRouteStatuses)
```
bowei: Don't we want to collect the errors and then merge them?

pkg/controller/controller.go
```diff
@@ -457,6 +455,22 @@ func (c *Controller) updateGatewayRemoveFinalizer(ctx context.Context, namespace
 	})
 }
 
+// updateGatewayStatus updates the Gateway status in the API server if it has changed.
+func updateGatewayStatus(ctx context.Context, c *Controller, gateway *gatewayv1.Gateway, newGW *gatewayv1.Gateway, listenerStatuses []gatewayv1.ListenerStatus, syncErr error) error {
+	setGatewayConditions(newGW, listenerStatuses, syncErr)
```
bowei, line 457: So it will be better for this func to DeepCopy the GW that is passed in, otherwise this func has a side effect as well as I/O. At least the function comment should indicate that we are doing both, because this may be hard to keep track of later.

pkg/infra/envoy/proxy.go
```diff
@@ -137,14 +144,18 @@ func (r *ResourceManager) ensureConfigMap(ctx context.Context) error {
...
 func (r *ResourceManager) ensureDeployment(ctx context.Context) error {
 	logger := klog.FromContext(ctx)
 
 	deployment := r.renderDeployment()
-	_, err := r.client.AppsV1().Deployments(deployment.Namespace).Get(ctx, deployment.Name, metav1.GetOptions{})
+	dep, err := r.client.AppsV1().Deployments(deployment.Namespace).Get(ctx, deployment.Name, metav1.GetOptions{})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			_, err = r.client.AppsV1().Deployments(deployment.Namespace).Create(ctx, deployment, metav1.CreateOptions{})
+			dep, err = r.client.AppsV1().Deployments(deployment.Namespace).Create(ctx, deployment, metav1.CreateOptions{})
```
bowei, line 151: you should put a log here that the deployment is being created

pkg/infra/envoy/proxy.go
bowei: Let's add a comment here so we know that we are doing this explicitly:

// If the Deployment was just created, we will highly likely immediately fail
// here and return.

pkg/infra/envoy/proxy.go
```diff
@@ -153,13 +164,19 @@ func (r *ResourceManager) ensureDeployment(ctx context.Context) error {
...
+			return nil
+		}
 	}
-	logger.Info("Envoy proxy deployment is ready!")
-	return nil
+
+	return fmt.Errorf("envoy deployment %s is not available yet", deployment.Name)
 }
 
+// ensureService ensures that the Service for the Envoy proxy exists and has a ClusterIP assigned.
```
bowei, line 172: I just think the Ensuring all the Envoy stuff should be in a separate controller  (read: different struct/ source file) and the GW controller should check that the Envoy is ok. If it is not, then GW controller should just bail out.

This will make the code easier to understand.

Let's do this as a follow on PR.

pkg/infra/envoy/resourcerender.go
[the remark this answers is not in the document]
bowei: Why not make setting extra labels a command line flag if it's what we want? More importantly, given no one has asked for it, we should leave this feature out for now? Seems very easy to add as a small follow up vs maintaining this invariant/behavior hidden in the choices we make in the code. IMO that is not going to work very well long-term.

bowei (commented):
I am ok with this fix but there are some comments.

There is a follow up on splitting out the Envoy deploy from the GW controller. We should decouple those two things -- it will make the controller behave better and be easier to maintain long-term.

pkg/infra/envoy/resourcerender.go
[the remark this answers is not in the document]
haiyanmeng: https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/gateway-infrastructure.go verifies the propagation of metadata from Gateway infrastructure to generated components. However, this test is core. I worked on this PR before we decided to only focus on core tests.

pkg/controller/controller.go
haiyanmeng: newGW started as a deepcopy of gateway in :
```
	newGW := gateway.DeepCopy()

	// Ensure Envoy proxy deployment and service exist.
	rm := envoy.NewResourceManager(c.core.client, gateway, c.envoyImage, c.agenticIdentityTrustDomain)
	proxyIP, err := rm.EnsureProxyExist(ctx)
	if err != nil {
		_ = updateGatewayStatus(ctx, c, gateway, newGW, nil, err)
		return err
	}
```
I added a comment to `updateGatewayStatus` saying that it modified `newGW`.

pkg/infra/envoy/proxy.go
haiyanmeng: Done

pkg/infra/envoy/proxy.go
haiyanmeng: Done

pkg/infra/envoy/proxy.go
haiyanmeng: SG

pkg/controller/controller.go
haiyanmeng: Good point. Done

pkg/controller/controller.go
haiyanmeng: The controller is responsible for creating and managing the Envoy deployment on-demand for each Gateway resource it reconciles. Non-blocking wait seems better

pkg/controller/controller.go
bowei: My comment was about removing side-effects from the func. I know that it is a DeepCopy, but that was because I read the code as part of the code review. I think if we can make this more side effect free, it is better? You don't have to put an extra comment on the func in that case as well.

Seems like a better tradeoff to me?

If you want to modify the copied GW, do it outside this func:

```
...
setGatewayErrStatus(gw, err)
updateGatewayStatus(gw)
...
```

bowei: [minor] Also, we will want to render the syncErr into a non-internal error code message. If you pass the error directly, it may be non-sensical for the user.

pkg/controller/controller.go
```diff
@@ -457,6 +459,23 @@ func (c *Controller) updateGatewayRemoveFinalizer(ctx context.Context, namespace
 	})
 }
 
+// updateGatewayStatus updates the Gateway status in the API server if it has changed.
+// Note: this function modifies newGW.
+func updateGatewayStatus(ctx context.Context, c *Controller, gateway *gatewayv1.Gateway, newGW *gatewayv1.Gateway, listenerStatuses []gatewayv1.ListenerStatus, syncErr error) error {
+	setGatewayConditions(newGW, listenerStatuses, syncErr)
+	if !reflect.DeepEqual(gateway.Status, newGW.Status) {
```
bowei: return immediately instead

if reflect.DeepEqual(...) { return }

pkg/controller/controller.go
```diff
@@ -457,6 +459,23 @@ func (c *Controller) updateGatewayRemoveFinalizer(ctx context.Context, namespace
...
+		if _, statusErr := c.gateway.client.GatewayV1().Gateways(newGW.Namespace).UpdateStatus(ctx, newGW, metav1.UpdateOptions{}); statusErr != nil {
+			klog.Errorf("Failed to update gateway status: %v", statusErr)
+			if syncErr == nil {
```
bowei: I don't understand this line.

pkg/controller/controller.go
```diff
@@ -418,25 +419,26 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
...
+	var errs []error
+	if err := updateGatewayStatus(ctx, c, gateway, newGW, listenerStatuses, xdsErr); err != nil {
+		errs = append(errs, err)
+	}
+	if err := c.updateHTTPRouteStatuses(ctx, httpRouteStatuses); err != nil {
+		errs = append(errs, err)
 	}
-	return c.updateHTTPRouteStatuses(ctx, httpRouteStatuses)
+	if len(errs) > 0 {
+		return errors.Join(errs...)
```
bowei: You might want to check that errors.Join(nil) does the right thing, avoids the extra conditionals here.

pkg/infra/envoy/proxy.go
```diff
@@ -114,6 +119,9 @@ func (r *ResourceManager) ensureSA(ctx context.Context) error {
 	return nil
 }
 
+// ensureConfigMap ensures that the ConfigMap for the Envoy proxy exists.
+// Note that this function does not poll or wait; it creates the resource if not found
```
bowei: let's remove this extra Note. Because the functions were renamed from `wait...`
it's not as confusing anymore.

haiyanmeng: Done

pkg/controller/controller.go
```diff
@@ -400,6 +400,7 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
 	rm := envoy.NewResourceManager(c.core.client, gateway, c.envoyImage, c.agenticIdentityTrustDomain)
 	proxyIP, err := rm.EnsureProxyExist(ctx)
 	if err != nil {
+		_ = updateGatewayStatus(ctx, c, gateway, newGW, nil, err)
```
bowei: Do we want to throw away the error here?

pkg/controller/controller.go
haiyanmeng: From https://pkg.go.dev/errors#Join: `Join returns nil if every value in errs is nil.` So it is safe to remove the `if` condition.

pkg/controller/controller.go
haiyanmeng: The parameter `gateway` is the origin Gateway object, and will not be changed. 
The parameter `newGW` will be changed. 
`updateGatewayStatus` was called in multiple different places, which was why `setGatewayConditions` was inside `updateGatewayStatus`.

pkg/controller/controller.go
haiyanmeng: Done

pkg/controller/controller.go
haiyanmeng: This is to make sure: If syncErr is nil, return the status update error to trigger a requeue.

pkg/controller/controller.go
haiyanmeng: I agree this is confusion. 
I changed it to `return updateGatewayStatus(ctx, c, gateway, newGW, nil, err)`, and updated the func comment of `updateGatewayStatus` to make it clear that it returns the original syncErr if it is non-nil.

pkg/controller/controller.go
haiyanmeng: > [minor] Also, we will want to render the syncErr into a non-internal error code message. If you pass the error directly, it may be non-sensical for the user.

https://github.com/kubernetes-sigs/kube-agentic-networking/blob/bbeb9d74ceda30cc5f513372b12ca99049743489/pkg/controller/gateway.go#L119-L128
Is it not enough?

pkg/controller/controller.go
```diff
@@ -457,6 +455,32 @@ func (c *Controller) updateGatewayRemoveFinalizer(ctx context.Context, namespace
...
 
+// updateGatewayStatus updates the Gateway status in the API server if it has changed.
+// Note: this function modifies newGW.
+// It returns the original syncErr if it is non-nil. If syncErr is nil and the status
+// update fails, it returns the status update error.
+
+func updateGatewayStatus(ctx context.Context, c *Controller, gateway *gatewayv1.Gateway, newGW *gatewayv1.Gateway, listenerStatuses []gatewayv1.ListenerStatus, syncErr error) error {
+	setGatewayConditions(newGW, listenerStatuses, syncErr)
+
+	if reflect.DeepEqual(gateway.Status, newGW.Status) {
```
bowei, line 459: We should return nil. 

I think it will be more confusing to make this pass through the error from syncErr. Let's say the caller of this got "Error: invalid fd" (as an example).

They may have passed in "Error: invalid fd" as the syncErr OR we got the error from inside this method -- this will be confusing. I would rather have:


syncErr := ...
...
updateErr := updateGatewayStatus(..., syncErr)

// do something with syncErr and updateErr -- you have kept them apart so
// you can easily tell which is which.

pkg/controller/controller.go
```diff
@@ -457,6 +455,32 @@ func (c *Controller) updateGatewayRemoveFinalizer(ctx context.Context, namespace
...
+		// If syncErr is nil, return the status update error to trigger a requeue.
+		// If syncErr is not nil, we return it instead to avoid masking the primary
+		// failure that caused the reconciliation to fail.
+		if syncErr == nil {
+			return statusErr
+		}
+	} else {
+		klog.FromContext(ctx).Info("Updated gateway status")
+	}
+	return syncErr
```
bowei: See comment above.

pkg/controller/controller.go
haiyanmeng: Done

pkg/controller/controller.go
haiyanmeng: Done

pkg/controller/controller.go
```diff
@@ -470,6 +451,22 @@ func (c *Controller) updateGatewayRemoveFinalizer(ctx context.Context, namespace
...
+// Note: this function modifies newGW.
+func updateGatewayStatus(ctx context.Context, c *Controller, gateway *gatewayv1.Gateway, newGW *gatewayv1.Gateway, listenerStatuses []gatewayv1.ListenerStatus, syncErr error) error {
+	setGatewayConditions(newGW, listenerStatuses, syncErr)
+
+	if reflect.DeepEqual(gateway.Status, newGW.Status) {
+		return nil
+	}
+
+	if _, statusErr := c.gateway.client.GatewayV1().Gateways(newGW.Namespace).UpdateStatus(ctx, newGW, metav1.UpdateOptions{}); statusErr != nil {
+		return fmt.Errorf("failed to update gateway status: %w", statusErr)
```
bowei, line 465: You should actually log error on this path if you are going to log on the success path.  Fix this in a follow on PR

pkg/controller/controller_test.go
```diff
@@ -342,3 +352,95 @@ func ptrGatewayKind(kind string) *gatewayv1.Kind {
...
+			client: fakeK8sClient,
+		},
+		gatewayqueue: queue,
+	}
+
+	err := c.syncGateway(context.Background(), gwNamespace+"/"+gwName)
+	if err == nil {
+		t.Fatal("expected syncGateway to return an error, but it returned nil")
+	}
+	expectedErr := "failed to create envoy serviceaccount: forced error"
```
bowei: remove this it will be fragile

pkg/infra/envoy/proxy.go
```diff
@@ -72,6 +70,9 @@ func proxyName(namespace, name string) string {
 
 // EnsureProxyExist ensures that the Envoy proxy deployment, service, and other resources exist and are ready.
 // It returns the ClusterIP of the proxy service.
+// Note that this function does not poll or wait for resources to become available;
```
bowei: I think we should delete these comments (it was a comment from earlier review?)

pkg/infra/envoy/proxy.go
```diff
@@ -137,14 +142,19 @@ func (r *ResourceManager) ensureConfigMap(ctx context.Context) error {
 	return nil
 }
 
+// ensureDeployment ensures that the Envoy deployment exists and is available.
+// Note that this function does not poll or wait for the deployment to become available;
```
bowei: see previous comment

pkg/infra/envoy/proxy_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	t.Run("deployment not found, creates and returns error (not ready)", func(t *testing.T) {
+		client := fake.NewClientset()
+		rm := NewResourceManager(client, gw, "envoy-image", "cluster.local")
+
+		err := rm.ensureDeployment(context.Background())
+		if err == nil {
+			t.Fatal("expected error as deployment is not available yet")
+		}
+		if err.Error() != "envoy deployment "+nodeID+" is not available yet" {
```
bowei: remove string comparison it will be fragile

pkg/infra/envoy/proxy_test.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+		client := fake.NewClientset(svc)
+		rm := NewResourceManager(client, gw, "envoy-image", "cluster.local")
+
+		ip, err := rm.ensureService(context.Background())
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		if ip != "10.0.0.1" {
+			t.Errorf("expected IP 10.0.0.1, got %s", ip)
```
bowei: You might want to make all of test error message in the standard go test format 


"Blah() = xx, want x"

pkg/infra/envoy/resourcerender.go
```diff
@@ -108,6 +108,33 @@ func generateSdsConfig(tmpl, sdsConfigName, trustDomain string) (string, error)
...
+	}
+	return labels
+}
+
+// getAnnotations converts the Gateway Infrastructure annotations map to a standard map[string]string
+// required by Kubernetes ObjectMeta. This conversion is necessary because Gateway API uses distinct
+// types for annotation keys and values, which are not directly assignable to plain strings in Go maps.
+func getAnnotations(infraAnnotations map[gatewayv1.AnnotationKey]gatewayv1.AnnotationValue) map[string]string {
+	if len(infraAnnotations) == 0 {
+		return nil
```
bowei: won't nil map blow up?

pkg/infra/envoy/resourcerender.go
```diff
@@ -144,13 +171,18 @@ func (r *ResourceManager) renderConfigMap() (*corev1.ConfigMap, error) {
 
 func (r *ResourceManager) renderDeployment() *appsv1.Deployment {
 	replicas := int32(1)
+	var infraAnnotations map[gatewayv1.AnnotationKey]gatewayv1.AnnotationValue
+	var infraLabels map[gatewayv1.LabelKey]gatewayv1.LabelValue
+	if r.gw.Spec.Infrastructure != nil {
+		infraAnnotations = r.gw.Spec.Infrastructure.Annotations
```
bowei, line 174: dont need to name these variables?

pkg/infra/envoy/resourcerender.go
```diff
@@ -162,10 +194,12 @@ func (r *ResourceManager) renderDeployment() *appsv1.Deployment {
 			},
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
-					Labels: map[string]string{
-						"app":                      r.nodeID,
-						constants.GatewayNameLabel: r.gw.Name,
-					},
+					Labels: func() map[string]string {
+						l := getLabels(r.gw.Name, infraLabels)
```
bowei, line 195: isn't infraLabels just empty?

pkg/infra/envoy/resourcerender.go
```diff
@@ -144,13 +171,18 @@ func (r *ResourceManager) renderConfigMap() (*corev1.ConfigMap, error) {
...
 	return &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      r.nodeID,
-			Namespace: r.namespace,
-			Labels: map[string]string{
-				constants.GatewayNameLabel: r.gw.Name,
-			},
+			Name:            r.nodeID,
+			Namespace:       r.namespace,
+			Labels:          getLabels(r.gw.Name, infraLabels),
```
bowei, line 181: isn't infraLabels nil?

pkg/infra/envoy/resourcerender.go
```diff
@@ -294,13 +328,18 @@ func (r *ResourceManager) renderService() *corev1.Service {
...
 	return &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      r.nodeID,
-			Namespace: r.namespace,
-			Labels: map[string]string{
-				constants.GatewayNameLabel: r.gw.Name,
-			},
+			Name:            r.nodeID,
+			Namespace:       r.namespace,
+			Labels:          getLabels(r.gw.Name, infraLabels),
```
bowei, line 338: see above

pkg/infra/envoy/resourcerender.go
```diff
@@ -314,13 +353,18 @@ func (r *ResourceManager) renderService() *corev1.Service {
 }
 
 func (r *ResourceManager) renderServiceAccount() *corev1.ServiceAccount {
+	var infraAnnotations map[gatewayv1.AnnotationKey]gatewayv1.AnnotationValue
+	var infraLabels map[gatewayv1.LabelKey]gatewayv1.LabelValue
```
bowei, line 354: isn't this going to be nil and then the func calls below don't do anything?

pkg/controller/controller.go
haiyanmeng: Done

pkg/controller/controller_test.go
haiyanmeng: Done

pkg/infra/envoy/proxy.go
haiyanmeng: There was an earlier comment on this. I missed this one. Good catch. Removed.

pkg/infra/envoy/proxy.go
haiyanmeng: Removed.

pkg/infra/envoy/proxy_test.go
haiyanmeng: Done

pkg/infra/envoy/resourcerender.go
haiyanmeng: It will only blow up if any future code tries to write to it directly. 

Changed it to an empty map to be safe.

pkg/infra/envoy/resourcerender.go
haiyanmeng: We need the check `r.gw.Spec.Infrastructure != nil` on line 176. 

In the Gateway API gatewayv1.GatewaySpec, the Infrastructure field is a pointer (*GatewayInfrastructure) and is optional.

If a user creates a Gateway without specifying the infrastructure field, r.gw.Spec.Infrastructure will be nil. Without this check, attempting to access r.gw.Spec.Infrastructure.Annotations or r.gw.Spec.Infrastructure.Labels directly would cause a runtime panic (nil pointer dereference).

pkg/infra/envoy/resourcerender.go
haiyanmeng: Not always. `infraLabels` comes from `r.gw.Spec.Infrastructure.Labels`

pkg/infra/envoy/resourcerender.go
haiyanmeng: Not always. `infraLabels` comes from `r.gw.Spec.Infrastructure.Labels`

pkg/infra/envoy/resourcerender.go
haiyanmeng: Not always. `infraLabels` comes from `r.gw.Spec.Infrastructure.Labels`

pkg/infra/envoy/resourcerender.go
haiyanmeng: We need the check r.gw.Spec.Infrastructure != nil on line 358.

In the Gateway API gatewayv1.GatewaySpec, the Infrastructure field is a pointer (*GatewayInfrastructure) and is optional.

If a user creates a Gateway without specifying the infrastructure field, r.gw.Spec.Infrastructure will be nil. Without this check, attempting to access r.gw.Spec.Infrastructure.Annotations or r.gw.Spec.Infrastructure.Labels directly would cause a runtime panic (nil pointer dereference).

pkg/infra/envoy/proxy_test.go
haiyanmeng: Done

bowei (approved):

bowei:
/lgtm

Thanks for the revisions.

haiyanmeng:
> Thanks for the revisions.

Thanks for the thorough reviews.

pkg/constants/envoynames.go
```diff
@@ -38,7 +38,8 @@ const (
 	EnvoyBootstrapMountPath = "/etc/envoy/bootstrap"
 
 	// GatewayNameLabel is the label key used to identify resources associated with a specific Gateway.
-	GatewayNameLabel = ProjectName + "/gateway-name"
+	// This label was formally introduced in https://gateway-api.sigs.k8s.io/geps/gep-1762/#automated-deployments.
+	GatewayNameLabel = "gateway.networking.k8s.io/gateway-name"
```
guicassolato, line 42: I believe this change may have broken the QuickStart: https://github.com/kubernetes-sigs/kube-agentic-networking/blob/eef9db463ec3a8d97eeee10d83e57f28b33aecb5/site-src/guides/quickstart/run-quickstart.sh#L245-L246

haiyanmeng: Good catch. I am currently testing our quickstart, and will report back my testing results.

haiyanmeng: It does break:
```
[INFO] Waiting for Envoy proxy deployment to be created...
[ERROR] Timed out waiting for Envoy proxy deployment to be created.
make: *** [Makefile:184: quickstart-gemini] Error 1
```

I will send out a PR to fix this today.

guicassolato: Thanks!

Please make sure to include: https://github.com/kubernetes-sigs/kube-agentic-networking/blob/eef9db463ec3a8d97eeee10d83e57f28b33aecb5/dev/ci/run-e2e.sh#L179

haiyanmeng: https://github.com/kubernetes-sigs/kube-agentic-networking/pull/251 should fix this.

PR #211, opened 2026-03-27
Various build fixes

bowei:
* Run golint as current user (prevent creating files as root)
* Use valid YAML lint version
* mkdocs/generate.sh should use a mktemp directory
* Ignore .dotfiles by default in boilerplate
* Limit Python lint to directories containing Python source

/kind cleanup

```release-note
NONE
```


LiorLieberman:
>Run golint as current user (prevent creating files as root)

Where is that change?

Issue #217, opened 2026-04-02
AccessPolicy RBAC denies requests incorrectly


LiorLieberman:
**What happened**:
Empty tool list inside authorization struct in AccessPolicy needs to be implemented with "allow nothing" vs deny. Also, we are likely need to consider the current approach of adding different RBACs vs having one rbac for multiple policies targeting the same backend/gateway

**What you expected to happen**:
Empty allow will serve as "Deny-by-default" and additional policies will gradually allow it.

**How to reproduce it (as minimally and precisely as possible)**:
Apply two policies targeting the same backend, one with an empy `authorization.tools` list and the other one with one tool. The expectation is that other tool is allowed and the rest is denied

**Anything else we need to know?**:

/cc @haiyanmeng

Issue #214, opened 2026-04-01
Unexpected tool access deny when InlineTools and ExternalAuth are used together


haiyanmeng:
**What happened**:
When InlineTools and ExternalAuth are used together to control the tool access of an agent, tool access was denied unexpectedly.

**What you expected to happen**:
The prompt `Read the wiki content of the GitHub repo kubernetes-sigs/kube-agentic-networking` (this calls the tool `read_wiki_contents`) should be allowed.

**How to reproduce it (as minimally and precisely as possible)**:
After running https://kube-agentic-networking.sigs.k8s.io/guides/external-auth-quickstart/#quickstart, I modified the XAccessPolicy object to be (it adds an `InlineTools` rule):
```
apiVersion: agentic.prototype.x-k8s.io/v0alpha0
kind: XAccessPolicy
metadata:
  name: auth-policy-remote-mcp
  namespace: quickstart-ns
spec:
  targetRefs:
    - group: agentic.prototype.x-k8s.io
      kind: XBackend
      name: remote-mcp-backend
  rules:
    - name: inline-tools-for-adk-agent-sa
      source:
        type: ServiceAccount
        serviceAccount:
          name: adk-agent-sa
          namespace: quickstart-ns
      authorization:
        type: InlineTools
        tools: [ "read_wiki_structure" ]
    - name: ext-authorizer-for-adk-agent-sa
      source:
        type: ServiceAccount
        serviceAccount:
          name: adk-agent-sa
          namespace: quickstart-ns
      authorization:
        type: ExternalAuth
        externalAuth:
          backendRef:
            kind: Service
            name: authorino-authorino-authorization
            namespace: authorino-operator
            port: 50051
          protocol: GRPC
          grpc: {}
```
Then I used the prompt `Read the wiki content of the GitHub repo kubernetes-sigs/kube-agentic-networking` (this calls the tool `read_wiki_contents`) in the agent and got 403 Access Denied error. 

**Anything else we need to know?**:

The generated Envoy config for the backend is:
```
             {
              "name": "quickstart-ns-remote-mcp-backend",
              "weight": 1,
              "typed_per_filter_config": {
               "envoy.filters.http.rbac.backend_level_1": {
                "@type": "type.googleapis.com/envoy.extensions.filters.http.rbac.v3.RBACPerRoute",
                "rbac": {
                 "rules": {
                  "policies": {
                   "inline-tools-for-adk-agent-sa": {
                    "permissions": [
                     {
[...]

Issue #55, opened 2026-01-26
Allow AccessPolicy to target Gateway objects


haiyanmeng:
**What would you like to be added**:
Currently, the [AccessPolicy](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/main/docs/proposals/0008-ToolAuthAPI.md#accesspolicy-crd) resource proposed in the kube-agentic-networking project only allows targeting [Backends](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/main/docs/proposals/0008-ToolAuthAPI.md#backend-crd).

I would like to allow `AccessPolicy` to target `Gateway` objects, in addition to `Backend` objects.

**Why this is needed**:

This enhancement will make the `AccessPolicy` API more scalable for scenarios where a given policy needs to be enforced for all the traffic managed by a `Gateway` object regardless of what the backend is.

Note: this enhancement was discussed in the OSS community meeting on Jan 22, 2026.

PR #220, opened 2026-04-03
Enhance the evaluation rules of AccessPolicy objects

haiyanmeng:
**What type of PR is this?**

/kind feature

The original evaluation rules had several issues:

1. When an object is targeted by multiple InlineTools AccessPolicy objects, the request is only allowed if all the AccessPolicy objects allow the request.

   * Expected behavior: the request should be allowed if any of the AccessPolicy objects allows the request.

   * Rationale: Be consistent with prior art like Istio.

2. The evaluation does not specify the evaluation order when both InlineTools and ExternalAuth rules exist for a target object.

   * Expected behavior: When both InlineTools and ExternalAuth rules exist for a target object, ExternalAuth rules should be evaluated first.

   * Rationale: efficiency + be consistent with prior art like Istio.

In addition, to make it easy for users to understand why a tool access is allowed or denied, we will disallow the combination of InlineTools and ExternalAuth in the same AccessPolicy.

Addresses [#217](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/217)

```release-note
[Breaking Change] When an object is targeted by multiple InlineTools AccessPolicy objects, the request is allowed if any of the AccessPolicy objects allows the request.
[Breaking Change] Combining InlineTools and ExternalAuth in the same AccessPolicy is not allowed.
```

### DIFF
--- a/docs/proposals/0059-AccessPolicyTargetRefs.md
+++ b/docs/proposals/0059-AccessPolicyTargetRefs.md
@@ -51,9 +51,9 @@ The following table summarizes the evaluation results for different combinations
 | Gateway ExternalAuth | Gateway InlineTools | Backend ExternalAuth | Backend InlineTools | Final Result |
 | :--- | :--- | :--- | :--- | :--- |
 | **Deny** (Any denies) | - | - | - | **Deny** |
-| **Allow** (All allow) | **Deny** (All deny) | - | - | **Deny** |
+| **Allow** (All allow) | **Deny** (none allows) | - | - | **Deny** |
 | **Allow** (All allow) | **Allow** (At least one allows) | **Deny** (Any denies) | - | **Deny** |
-| **Allow** (All allow) | **Allow** (At least one allows) | **Allow** (All allow) | **Deny** (All deny) | **Deny** |
+| **Allow** (All allow) | **Allow** (At least one allows) | **Allow** (All allow) | **Deny** (none allows) | **Deny** |
 | **Allow** (All allow) | **Allow** (At least one allows) | **Allow** (All allow) | **Allow** (At least one allows) | **Allow** |
 
 ## Example
@@ -154,7 +154,7 @@ When a request comes to `payment-service` through `prod-gateway`:
     - *Note: Evaluation order among these policies does not matter.*
 - **InlineTools Policies**: `gateway-policy-inline-tools-1` and `gateway-policy-inline-tools-2` are evaluated.
     - If **any** allows, proceed to Backend level.
-    - If **all** deny, the request is rejected immediately.
+    - If **none** allows, the request is rejected immediately.
     - *Note: Evaluation order among these policies does not matter.*
 
 #### 2. Backend Level Checks
@@ -164,7 +164,7 @@ When a request comes to `payment-service` through `prod-gateway`:
     - *Note: Evaluation order among these policies does not matter.*
 - **InlineTools Policies**: `backend-policy-inline-tools-1` and `backend-policy-inline-tools-2` are evaluated.
     - If **any** allows, the request is allowed.
-    - If **all** deny, the request is denied.
+    - If **none** allows, the request is denied.
     - *Note: Evaluation order among these policies does not matter.*
 
 
--- a/docs/proposals/0059-AccessPolicyTargetRefs.md
+++ b/docs/proposals/0059-AccessPolicyTargetRefs.md
@@ -8,16 +8,41 @@ Currently, the [AccessPolicy](https://github.com/kubernetes-sigs/kube-agentic-ne
 
 This has scalability issues when a given Tool Authorization policy needs to be enforced for all the traffic managed by a [Gateway](https://gateway-api.sigs.k8s.io/api-types/gateway/) object.
 
-This proposal allows `AccessPolicy` to target `Gateway` objects, in addition to `Backend` objects with following restrictions:
+This proposal allows `AccessPolicy` to target `Gateway` objects, in addition to `Backend` objects with the following rules and restrictions:
 
-1. A single `AccessPolicy` object targeting a Gateway and a Backend at the same time is NOT allowed.
+## Restrictions and Evaluation Order
 
-1. It is allowed to have `AccessPolicy` objects targeting a `Gateway` object and `AccessPolicy` objects targeting a `Backend` object behind the `Gateway` object. In this case, the `AccessPolicy` objects targeting the `Gateway` object will be evaluated first. Among the `AccessPolicy` objects targeting the `Gateway` object, the [ExternalAuth-type](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/cf8c85b85d067657a5dce7b87270f8099f1e302c/api/v0alpha0/accesspolicy_types.go#L169) authorization rules matching the request, if exist, will be evaluated first, the request will be denied if the evaluation result of any of the ExternalAuth-type authorization rules is `deny`. If there are no [InlineTools-type](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/cf8c85b85d067657a5dce7b87270f8099f1e302c/api/v0alpha0/accesspolicy_types.go#L165) authorization rules matching the request, allow the request. If any InlineTools-type authorization rule matching the request exists, allow the request. Otherwise, deny the request.
+### Restrictions
+- **Mutually Exclusive Targets**: A single `AccessPolicy` object targeting a Gateway and a Backend at the same time is **NOT** allowed.
 
+### Coexistence and Evaluation Flow
+It is allowed to have `AccessPolicy` objects targeting a `Gateway` object and `AccessPolicy` objects targeting a `Backend` object behind the `Gateway` object. The evaluation follows a strict hierarchy:
 
-    * If the HTTP request is denied at the gateway-level evaluation, the `AccessPolicy` objects targeting the `Backend` object will NOT be evaluated.
+#### 1. Gateway-Level Evaluation (First)
+The `AccessPolicy` objects targeting the `Gateway` object are evaluated first.
 
-    * If the HTTP request is allowed at the gateway-level evaluation, the `AccessPolicy` objects targeting the `Backend` object will be evaluated. Among the `AccessPolicy` objects targeting the `Backend` object, the [ExternalAuth-type](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/cf8c85b85d067657a5dce7b87270f8099f1e302c/api/v0alpha0/accesspolicy_types.go#L169) authorization rules matching the request, if exist, will be evaluated first, the request will be denied if the evaluation result of any of the ExternalAuth-type authorization rules is `deny`. If there are no [InlineTools-type](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/cf8c85b85d067657a5dce7b87270f8099f1e302c/api/v0alpha0/accesspolicy_types.go#L165) authorization rules matching the request, allow the request. If any InlineTools-type authorization rule matching the request exists, allow the request. Otherwise, deny the request.
+- **[ExternalAuth-type](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/cf8c85b85d067657a5dce7b87270f8099f1e302c/api/v0alpha0/accesspolicy_types.go#L169) Rules**:
+    - Evaluated first.
+    - If **any** ExternalAuth rule matching the request returns `deny`, the request is denied immediately.
+- **[InlineTools-type](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/cf8c85b85d067657a5dce7b87270f8099f1e302c/api/v0alpha0/accesspolicy_types.go#L165) Rules**:
+    - Evaluated only if all ExternalAuth rules (if any) allowed the request.
+    - If there are **no** InlineTools rules matching the request, the request is allowed.
+    - If **any** InlineTools rule matching the request exists and allows it, the request is allowed.
+    - Otherwise, the request is denied.
+
+If the request is denied at the gateway level, evaluation stops, and Backend-level policies are **NOT** evaluated.
+
+#### 2. Backend-Level Evaluation (Second)
+If the request is allowed at the gateway level, the `AccessPolicy` objects targeting the `Backend` object are evaluated using the same logic:
+
+- **[ExternalAuth-type](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/cf8c85b85d067657a5dce7b87270f8099f1e302c/api/v0alpha0/accesspolicy_types.go#L169) Rules**:
+    - Evaluated first.
+    - If **any** ExternalAuth rule matching the request returns `deny`, the request is denied immediately.
+- **[InlineTools-type](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/cf8c85b85d067657a5dce7b87270f8099f1e302c/api/v0alpha0/accesspolicy_types.go#L165) Rules**:
+    - Evaluated only if all ExternalAuth rules (if any) allowed the request.
+    - If there are **no** InlineTools rules matching the request, the request is allowed.
+    - If **any** InlineTools rule matching the request exists and allows it, the request is allowed.
+    - Otherwise, the request is denied.
 
 ## Summary Table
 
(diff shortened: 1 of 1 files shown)


guicassolato:
It's a bank holiday where I live so I will only be back on Tuesday. Therefore, I couldn't give this PR the thorough review it deserves yet, but here are my 2¢ based on what I saw at a glance:

1. The order of evaluation/enforcement of the rules should be the order of declaration - no obscure behaviour like `ExternalAuth` goes first, `InlineTools` next. This is what I think makes more sense for users, regardless of what implementation _X_ does. It's simple and straightforward.
2. There's no reason not to allow `ExternalAuth` and `InlineTools` in the same object – if they can be in different objects, so they should be allowed to declare in the same one. Please, check out as well these [5 reasons](https://github.com/kubernetes-sigs/kube-agentic-networking/pull/151#issuecomment-4053842748) why NOT to move `ExternalAuth` to the top level of the AccessPolicy, which effectively is what not allowing using `ExternalAuth` and `InlineTools` together is trying to do, I suppose.
3. ALL RULES must allow access for access to be granted, not AT LEAST ONE RULE, otherwise there may be a rogue policy that _grants_ access while the user struggles to figure out why another policy that _denies_ access seems not to be working as expected. It is the safe approach. Therefore, in bullet point number 2 of [this comment](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/214#issuecomment-4175108223),  I am of the opinion that the observed behaviour is in fact the correct one.

As I mentioned on [Slack](https://kubernetes.slack.com/archives/C09P6KS6EQZ/p1775221535411809?thread_ts=1775217898.607459&cid=C09P6KS6EQZ), I think it's good a thing that we're trying to fill some definition gaps. My first impression nevertheless was that some of the new definitions proposed here feel like we may be overthinking it a little. This only means I think we should give stakeholders a chance of forming a _second_ impression, hence my suggestion to hold at least a second round of discussion and my commitment to review the PR more thoroughly, trying to better understand the other POVs.

robscott:
> > > ALL RULES must allow access for access to be granted, not AT LEAST ONE RULE, otherwise there may be a rogue policy that grants access while the user struggles to figure out why another policy that denies access seems not to be working as expected. It is the safe approach. Therefore, in bullet point number 2 of [#214 (comment)](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/214#issuecomment-4175108223), I am of the opinion that the observed behaviour is in fact the correct one.
> > 
> > 
> > Under that framework, don't you have the inverse problem where a rogue policy can block traffic to ~any endpoint?
> 
> Yes, and this is a good thing. That’s what meant by the safe approach.

Just to make sure I understand the conversation here, let me provide an example:

policy 1:
- rule a: identity foo can talk to the github tool

policy 2:
- rule b: identity foo can talk to the gitlab tool

Taking this guideline literally, it seems like foo would not be able to talk to github or gitlab. That's because although policy 1 allows foo to access github, it does not allow foo to access gitlab. With a similar situation for policy 2. Is that what you're intending to say @guicassolato? I'm assuming I'm misunderstanding here.

docs/proposals/0059-AccessPolicyTargetRefs.md
```diff
@@ -12,15 +12,24 @@ This proposal allows `AccessPolicy` to target `Gateway` objects, in addition to
...
+1. It is allowed to have `AccessPolicy` objects targeting a `Gateway` object and `AccessPolicy` objects targeting a `Backend` object behind the `Gateway` object. In this case, the `AccessPolicy` objects targeting the `Gateway` object will be evaluated first. Among the `AccessPolicy` objects targeting the `Gateway` object, the [ExternalAuth-type](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/cf8c85b85d067657a5dce7b87270f8099f1e302c/api/v0alpha0/accesspolicy_types.go#L169) authorization rules matching the request, if exist, will be evaluated first, the request will be denied if the evaluation result of any of the ExternalAuth-type authorization rules is `deny`. If there are no [InlineTools-type](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/cf8c85b85d067657a5dce7b87270f8099f1e302c/api/v0alpha0/accesspolicy_types.go#L165) authorization rules matching the request, allow the request. If any InlineTools-type authorization rule matching the request exists, allow the request. Otherwise, deny the request.
```
robscott: Because "no rules matching" and "any rules matching" cover 100% of all possible scenarios, the "Otherwise, deny" condition can never be reached. Suggesting a few changes to make this clearer:

```suggestion
1. Separate `AccessPolicy` resources can target a `Gateway` object and a `Backend` object behind that `Gateway` object. In this case, the `AccessPolicy` objects targeting the `Gateway` object will be evaluated first. Among the `AccessPolicy` objects that apply to a request, any ExternalAuth-type authorization rules will be evaluated first. The request will be denied if the evaluation result of any of the ExternalAuth-type authorization rules is `deny`. If any InlineTools-type authorization rule matching the request exists, the request will be allowed. Otherwise, the request will be denied.
```

docs/proposals/0059-AccessPolicyTargetRefs.md
```diff
@@ -8,19 +8,53 @@ Currently, the [AccessPolicy](https://github.com/kubernetes-sigs/kube-agentic-ne
...
 
-        * if any of the `AccessPolicy` objects targeting the `Backend` object denies the access, the HTTP request will be denied.
+- **[ExternalAuth-type](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/cf8c85b85d067657a5dce7b87270f8099f1e302c/api/v0alpha0/accesspolicy_types.go#L169) Rules**:
+    - Evaluated first.
+    - If **any** ExternalAuth rule matching the request returns `deny`, the request is denied immediately.
+- **[InlineTools-type](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/cf8c85b85d067657a5dce7b87270f8099f1e302c/api/v0alpha0/accesspolicy_types.go#L165) Rules**:
+    - Evaluated only if all ExternalAuth rules (if any) allowed the request.
+    - If there are **no** InlineTools rules matching the request, the request is allowed.
+    - If **any** InlineTools rule matching the request exists and allows it, the request is allowed.
+    - Otherwise, the request is denied.
```
guicassolato: Although I'd still prefer the rules to be evaluated in the order they appear in the policy, I think this is an acceptable compromise if it's to keep the [deny-overrides behaviour](https://github.com/kubernetes-sigs/kube-agentic-networking/pull/220#issuecomment-4197844376).

For the reference implementation:

I think this will require configuring two RBAC filters - before and after ExternalAuth. So the config goes:
1. Gateway-level RBAC filter with shadow rules for ExternalAuth
2. Gateway-level Ext_authz filter
3. Gateway-level RBAC filter with normal rules for InlineTools
4. Backend-level RBAC filter with shadow rules for ExternalAuth
5. Backend-level Ext_authz filter
6. Backend-level RBAC filter with normal rules for InlineTools

docs/proposals/0059-AccessPolicyTargetRefs.md
```diff
@@ -8,19 +8,53 @@ Currently, the [AccessPolicy](https://github.com/kubernetes-sigs/kube-agentic-ne
...
+
+## Summary Table
+
+The following table summarizes the evaluation results for different combinations of policy checks. Note that evaluation short-circuits as soon as a definitive Deny or Allow is reached.
+
+| Gateway ExternalAuth | Gateway InlineTools | Backend ExternalAuth | Backend InlineTools | Final Result |
+| :--- | :--- | :--- | :--- | :--- |
+| **Deny** (Any denies) | - | - | - | **Deny** |
+| **Allow** (All allow) | **Deny** (All deny) | - | - | **Deny** |
```
guicassolato: How can you deny with `InlineTools` today?

docs/proposals/0059-AccessPolicyTargetRefs.md
```diff
@@ -61,39 +101,72 @@ graph TD
...
+    - *Note: Evaluation order among these policies does not matter.*
+
+#### 2. Backend Level Checks
+- **ExternalAuth Policies**: `backend-policy-external-auth-1` and `backend-policy-external-auth-2` are evaluated.
+    - If **any** denies, the request is rejected immediately.
+    - If **all** allow, proceed.
+    - *Note: Evaluation order among these policies does not matter.*
+- **InlineTools Policies**: `backend-policy-inline-tools-1` and `backend-policy-inline-tools-2` are evaluated.
+    - If **any** allows, the request is allowed.
+    - If **all** deny, the request is denied.
```
guicassolato:
```suggestion
    - If **none** allow, the request is denied.
```

docs/proposals/0059-AccessPolicyTargetRefs.md
```diff
@@ -126,3 +201,102 @@ Currently, the `InlineTools` type of [AuthorizationRule](https://github.com/kube
...
+
+Linkerd's proxy makes an authorization decision for every incoming request using these checks in order:
+
+1. Check for a `Server` resource:
+
+    * NO `Server` exists: The proxy uses the Default Policy (see below).
+
+    * YES `Server` exists: Proceed to step 2.
+
+1. Check for an `HTTPRoute`:
```
guicassolato: nit:

```suggestion
2. Check for an `HTTPRoute`:
```

docs/proposals/0059-AccessPolicyTargetRefs.md
```diff
@@ -61,39 +101,72 @@ graph TD
...
-    *   If (and only if) all Gateway policies allow the request, `backend-policy-admin` is evaluated.
-    *   **Rule:** Backend policies must also allow the request.
+#### 1. Gateway Level Checks
+- **ExternalAuth Policies**: `gateway-policy-external-auth-1` and `gateway-policy-external-auth-2` are evaluated.
+    - If **any** denies, the request is rejected immediately.
+    - If **all** allow, proceed.
+    - *Note: Evaluation order among these policies does not matter.*
+- **InlineTools Policies**: `gateway-policy-inline-tools-1` and `gateway-policy-inline-tools-2` are evaluated.
+    - If **any** allows, proceed to Backend level.
+    - If **all** deny, the request is rejected immediately.
```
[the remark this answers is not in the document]
guicassolato: Except that the request is _not explicitly denied_ at the gateway level. Rather, it's _not allowed_, which can then lead to _implicitly denied_, or alternatively could lead to _deferred to backend level_.

I wonder if an implicit deny at the gateway level when there are still other policies to check at backend level is the behaviour users expect.

This behaviour is equivalent to the [**atomic overrides**](https://gateway-api.sigs.k8s.io/geps/gep-713/#merge-strategy-atomic-overrides) merge strategy described by GEP-713. AFAIK, it has no counterpart in Istio, it's the exact opposite of the behaviour implemented by Envoy Gateway (atomic defaults), and exists in Kuadrant as an opt-in feature.

docs/proposals/0059-AccessPolicyTargetRefs.md
```diff
@@ -61,39 +101,72 @@ graph TD
...
-    *   **Rule:** ALL Gateway policies must allow the request. If `gateway-policy-audit` denies it, the request is rejected immediately, and subsequent policies are skipped.
-
-2.  **Backend Level Checks:**
-    *   If (and only if) all Gateway policies allow the request, `backend-policy-admin` is evaluated.
-    *   **Rule:** Backend policies must also allow the request.
+#### 1. Gateway Level Checks
+- **ExternalAuth Policies**: `gateway-policy-external-auth-1` and `gateway-policy-external-auth-2` are evaluated.
+    - If **any** denies, the request is rejected immediately.
+    - If **all** allow, proceed.
+    - *Note: Evaluation order among these policies does not matter.*
```
howardjohn, line 154: Not sure if this "does not matter" is meant to mean "in the way its specified" or "to users"?

I think its probably important to users the order; usually if they are chaining, they are using external auth not just to give a `allow/deny` but to augment the request with additional information.

However I agree its a very challenging problem to solve if we allow multiple across resources 🙂 just that I would expect that eventually we see demand for ordering

haiyanmeng: > I think its probably important to users the order; usually if they are chaining, they are using external auth not just to give a allow/deny but to augment the request with additional information.

Good point. 

Before this PR, the ordering is: the policies with earlier creationTimestamp will be evaluated first. For the policies with the same creationTimestamp, the ones appearing first in alphabetical order by `{namespace}/{name}` will be evaluated first.  

Maybe we can keep this ordering for both ExternalAuth and ALLOW/InlineTools policies (strictly speaking, the evaluation order of ALLOW policies really does not matter). WDYT?

docs/proposals/0059-AccessPolicyTargetRefs.md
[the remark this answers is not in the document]
haiyanmeng: > I wonder if an implicit deny at the gateway level when there are still other policies to check at backend level is the behaviour users expect.

I agree that we need user feedback regarding this. 
Before we have user feedback, let us stick to `implicit deny`. WDYT?

@guicassolato  @mikemorris @howardjohn

markfisher1:
**This PR is solving real problems:**

- Clarifies ambiguous behavior
- Aligns with ecosystem patterns
- Moves toward a usable model

But we’re at a foundational layer.

Decisions made here will:

- Define how people reason about security
- Shape future API constraints
- Either enable or block advanced use cases

**Recommendation Before Merge**

- Define a formal evaluation contract (deterministic spec)
- Clarify ALLOW vs DENY responsibilities
- Reconsider composition restriction (InlineTools + ExternalAuth)
- Explicitly define Gateway vs Backend authority model
- Leave room for ordering + context evolution

If we get this right now, this becomes a reference model for agentic policy in Kubernetes.

If we rush it, we’ll be revisiting this with breaking changes in a few releases.

docs/proposals/0059-AccessPolicyTargetRefs.md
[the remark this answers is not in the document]
mikemorris: I think the trickiest part here is distinguishing between an explicitly configured empty set `tools: []` field versus an unset/omitted field (which is particularly annoying in Go implementations, and would necessitate using a pointer type).

IMO, [strict hierarchical evaluation](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/55#issuecomment-3818898257) (Gateway before Backend) means:
- `tools: [y]` on a Gateway-targeted policy evaluates to an explicit ALLOW for calling tool `y`, any applicable Backend-targeted policy is evaluated next.
- `tools: [x]` on a Gateway-targeted policy evaluates to an _implicit_ "allow no tools except x" - an attempt to call tool `y` is not allowed, so any Backend-targeted policy is irrelevant
- `tools: []` on a Gateway-targeted policy _implicitly_ allows no tool calls, any Backend-targeted policy is irrelevant

Critically, the "implicit allow nothing" is _not_ actually modeled as a DENY which would halt further evaluation and reject the request, it's more like implicitly adding a higher-level default cluster-wide "allow nothing". In Istio this is actually [created explicitly as an "allow-nothing" policy](https://istio.io/latest/docs/reference/config/security/authorization-policy/), whereas NetworkPolicy enforces this implicitly when an ALLOW rule is created (similar to what's being proposed here).

```
kind: XAccessPolicy
spec:
  targetRefs:
  - kind: Gateway
  rules: []
```
^ (or similarly, the lack of any AccessPolicy targeting the Gateway) _does not specify_ any InlineTools ALLOW rules, and thus I would expect no implicit deny should exist and any Backend-targeted should be evaluated. Backend-targeted policies would also be evaluated when a Gateway-targeted AccessPolicy _does_ explicitly allow a requested tool.

This is a somewhat contrived case currently, but could be much more relevant if AccessPolicy eventually became an additional ruleset extension of [AuthorizationPolicy](https://gateway-api.sigs.k8s.io/geps/gep-3779/) (which would make broader L4/L7 rules omitting a `tools` stanza much more common) or added an audit/"shadow" mode (for logging/testing before enforcement) as below:
```
  rules:
  - authorization:
      type: InlineTools
      action: ALLOW
      mode: AUDIT | ENFORCED (default when mode field is omitted)
      tools: 
      - x
```
(I've always found AUDIT as a [peer enum value alongside ALLOW and DENY](https://istio.io/latest/docs/reference/config/security/authorization-policy/#AuthorizationPolicy-Action) as confusing, because the intent of the rule becomes less clear.)

If we add a DENY action, we should consider whether that would also trigger the implicit "allow nothing" layer or not (as a narrow DENY breaking all traffic might be unexpected).

docs/proposals/0059-AccessPolicyTargetRefs.md
mikemorris: I actually noticed this reading through https://github.com/kubernetes-sigs/kube-agentic-networking/blob/main/docs/proposals/0059-AccessPolicyTargetRefs.md and felt the timestamp behavior was a bit unexpected - in Gateway API timestamps are mostly used to [resolve conflicts](https://gateway-api.sigs.k8s.io/guides/api-design/#conflicts) during the evaluation of a resource by a controller (deciding whether to accept or reject it, such as if only a certain number of Gateway-targeted policies were allowed), and aren't typically relevant during evaluation/enforcement of a resource or policy by the dataplane at runtime.

> First, gateway-policy-audit is evaluated (earlier creation timestamp).
Next, gateway-policy-region is evaluated.

I can see how, as @howardjohn mentions, mutations which augment a request could be relevant for subsequent policy decisions, but I think a more explicit ordered configuration would be preferable there, and I'm not sure if the mutating scope extends beyond current expectations for AccessPolicy and starts looking more like https://github.com/kubernetes-sigs/wg-ai-gateway/blob/main/proposals/7-payload-processing.md, which does have specific expectations around ordering and request mutation?

haiyanmeng: @mikemorris , thanks for the feedback. Folk seems in favor of limiting the number of ExternalAuth policies/rules per target to 1 (see the GH [discussion](https://github.com/kubernetes-sigs/kube-agentic-networking/discussions/235#discussioncomment-16506417)).

I will update this PR after we finalize the GH dicussion regarding this.

PR #227, opened 2026-04-06
docs: Tool discovery proposal (XToolInventory CRD)

rubambiza:
## Summary

- Proposes optional, opt-in tool discovery for KAN: a controller that calls `tools/list` on MCP backends, validates schemas, and surfaces discovered tools in a dedicated `XToolInventory` CRD (1:1 with XBackend via `backendRef`)
- Enables policy validation against real backend state and operator visibility via `kubectl`
- Discovery is advisory (does not affect policy enforcement) and opt-in via hybrid creation model (operator creates XToolInventory explicitly, or controller auto-creates via annotation)

## Key design decisions

- **Separate CRD over XBackend.status**: Decouples discovery lifecycle from backend lifecycle; not every backend controller does discovery; XBackend shape is in flux
- **1:1 mapping**: One XToolInventory per XBackend, analogous to HTTPRoute-per-route
- **Hybrid creation**: Explicit creation or annotation-triggered auto-creation (common pattern: Service → EndpointSlice)
- **Freshness signals**: Honor a server-advertised `ttlMs` ([MCP SEP #2549](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2549)) as the primary refetch bound, clamped by `spec.maxPollInterval` to guard against excessively long TTLs; `tools/list_changed` as immediate invalidation; fall back to `spec.pollInterval` (default 5m) when neither is advertised
- **Validation scoped to Inline rules**: Cross-referencing discovered tools applies to `Inline` authorization (explicit enumeration), not `CEL` (dynamic matching)
- **v1alpha1 alignment**: Examples use `agentic.networking.x-k8s.io/v1alpha1` with method-based matching (`mcp.methods[].params[]`)

## Phasing

- **Phase 1**: In-cluster discovery (plain HTTP), XToolInventory CRD, schema validation, `list_changed` + `ttlMs` freshness handling, drift detection
- **Phase 2**: External backends with TLS/authentication via `credentialRef` on XToolInventory (deferred pending Backend spec stabilization)
- **Phase 3**: XAccessPolicy cross-referencing `params[]` against `XToolInventory.status.discoveredTools`

## Prior discussion

- Initial draft reviewed on fork PR (rubambiza/kube-agentic-networking#1) with feedback from @evaline-ju and @david-martin
- Major revision (2026-05-25) based on community feedback from @keithmattix (separate CRD, egress gateway routing) and @david-martin (v1alpha1 alignment, poll interval, two identities)
- Revision (2026-07-02) adding `ttlMs` freshness handling (SEP #2549) and Inline-scoped validation, based on feedback from @LiorLieberman, @david-martin, and @shachartal

## Community consensus points

- [x] Is XToolInventory with 1:1 backendRef the right granularity?
- [x] MCP client library: `github.com/mark3labs/mcp-go`?
- [x] Hybrid creation model (explicit + annotation-triggered) acceptable?
- [x] Freshness signals: `ttlMs` as a refetch bound (vs. on-access hint) + notifications — acceptable trade-off for a control-plane inventory?
- [x] Warning vs. rejection for unverified tools?

## Test plan

- [ ] Community review of revised API design (XToolInventory CRD)
- [ ] Alignment on consensus points above
- [ ] Validate implementability against current codebase structure

### DIFF
--- a/docs/proposals/XXXX-ToolDiscovery.md
+++ b/docs/proposals/XXXX-ToolDiscovery.md
@@ -1,0 +1,437 @@
+# Tool Discovery for KAN
+
+Date: 16th March 2026<br/>
+Authors: @rubambiza, @evaline-ju, @david-martin, @guicassolato<br/>
+Status: Provisional<br/>
+
+> [!CAUTION]
+> This API is in a provisional state. It is subject to change and should not be
+> implemented without first consulting the maintainers of the project.
+
+## Summary
+
+KAN has no mechanism to discover what tools an MCP backend exposes. Operators
+must manually enumerate tool names in `XAccessPolicy` rules, and there is no
+feedback loop when tools are added, removed, or misconfigured on backends.
+
+This proposal adds optional, opt-in tool discovery to KAN: a controller that
+connects to MCP backends, calls `tools/list`, validates the results, and
+surfaces discovered tools in `XBackend.status`. This enables policy validation
+against real backend state, operator visibility via `kubectl`, and a foundation
+for future capabilities like response filtering. Discovery is advisory (i.e.,
+it enhances observability but does not affect policy enforcement). Backends
+without discovery enabled continue to work exactly as they do today.
+
+## Non-Goals
+
+- **`tools/list` response filtering.** Filtering the `tools/list` response
+  based on caller identity (so agents only see tools they're authorized to call)
+  is a natural next step but would require a separate proposal.
+- **Tool prefixing / federation.** Disambiguating tools with the same name
+  across multiple backends (e.g., via prefix namespacing) is out of scope.
+- **Full tool schema storage in CRD status.** Storing `inputSchema` and
+  `outputSchema` in `XBackend.status` risks etcd bloat. Discovery validates
+  schemas but does not persist them.
+
+## Motivation
+
+### The Gap
+
+KAN's `XBackend` CRD is purely a routing target. It knows an MCP server's
+address and port, but nothing about what tools that server exposes:
+
+```go
+type MCPBackend struct {
+    ServiceName *string `json:"serviceName,omitempty"`
+    Hostname    *string `json:"hostname,omitempty"`
+    Port        int32   `json:"port"`
+    Path        string  `json:"path,omitempty"`
+}
+// No status.discoveredTools, no status.tools[].
+```
+
+This creates three concrete problems:
+
+**1. Stale access policies.** `XAccessPolicy` rules require operators to
+manually list tool names. If a backend adds, renames, or removes a tool, the
+policy becomes stale with no warning. The ToolAuthAPI proposal (0008) explicitly
+acknowledges this as a TODO.
+
+**2. No operator visibility.** There is no way to inspect what tools a backend
+exposes without directly calling `tools/list` on the backend. `kubectl get
+xbackend` tells you nothing about the backend's capabilities.
+
+**3. Invalid schemas propagate silently.** When a backend MCP server exposes
+tools with invalid JSON schemas, the problem is invisible until an agent tries
+to use the tools. For example, [Kuadrant/mcp-gateway issue #662](https://github.com/Kuadrant/mcp-gateway/issues/662) documents a case
+where a backend with an invalid `inputSchema` caused an agent to fail with a 400
+error (`"JSON schema is invalid. It must match JSON Schema draft 2020-12"`).
+Without validation at the discovery layer, broken backends silently poison the
+system.
+
+### Personas
+
+**Platform Engineer** — Configures `XBackend` and `XAccessPolicy` resources.
+Needs to know what tools each backend exposes to write correct policies, and
+wants warnings when policies reference tools that don't exist.
+
+**Tool Developer** — Deploys MCP servers behind KAN. Wants confidence that
+newly added or renamed tools are visible to the platform without manual CRD
+updates.
+
+### User Journeys
+
+**CUJ 1: Operator discovers tools on a new backend.**
+The platform engineer creates an `XBackend` pointing at an MCP server. Within
+the sync interval, `kubectl get xbackend my-backend -o yaml` shows
+`status.discoveredTools` populated with the backend's tools. No manual
+enumeration required.
+
+**CUJ 2: Operator writes a valid access policy.**
+The platform engineer writes an `XAccessPolicy` referencing tool names. The
+controller validates the tool names against `XBackend.status.discoveredTools`
+and sets a Warning condition on the policy if any tool name doesn't match.
+
+**CUJ 3: Backend adds a new tool.**
+A tool developer deploys a new version of their MCP server with an additional
+tool. The discovery controller detects the change (via `tools/list_changed`
+notification or the next poll) and updates `XBackend.status`. Existing
+`XAccessPolicy` resources that don't include the new tool are unaffected; the
+operator can add it when ready.
+
+**CUJ 4: Backend exposes an invalid tool.**
+A tool developer deploys an MCP server with a tool whose `inputSchema` is
+invalid JSON Schema. The discovery controller validates the schema, rejects the
+tool, and emits a Warning event on the `XBackend`. The invalid tool does not
+appear in `status.discoveredTools`, preventing it from being referenced in
+policies or reaching agents.
+
+## Context
+
+Three recent developments inform this proposal's design:
+
+**XBackend is in flux.** KAN is actively reconsidering XBackend's shape to
+align with the AI Gateway WG's Backend resource ([gateway-api PR #4488](https://github.com/kubernetes-sigs/gateway-api/pull/4488)). [Issue
+#161](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/161) tracks whether to drop `Service` as a target type; [issue #162](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/162) questions
+whether `path` belongs on XBackend or in protocol-specific options. This
+proposal's design is resilient to these changes: it extends `XBackend.status`
+(observed state), not `XBackend.spec`, so it works regardless of how the spec
+evolves.
+
+**MCP 2025-11-25 expanded the tool data model.** The MCP spec now includes
+`title` (human-readable display name), `outputSchema`, `icons`, pagination on
+`tools/list` via cursor, and mandatory JSON Schema 2020-12 validation for
+`inputSchema`. The `tools/list_changed` notification (available since
+2025-03-26) was not incorporated in the original proposal.
+
+**Kuadrant/mcp-gateway validates the approach.** Three developments in the MCP
+Gateway directly inform this proposal: `tools/list_changed` support ([PR #329](https://github.com/Kuadrant/mcp-gateway/pull/329))
+validates the hybrid polling + notification approach; [issue #662](https://github.com/Kuadrant/mcp-gateway/issues/662) (invalid
+schemas breaking downstream agents) motivates schema validation at the discovery
+layer; and the `MCPServerRegistration` CRD's evolution toward richer backend
+metadata ([PR #500](https://github.com/Kuadrant/mcp-gateway/pull/500)) confirms that a count-only status is insufficient for policy
+validation.
+
+## API
+
+### XBackend Status Extension
+
+```go
+type BackendStatus struct {
+    // Conditions describe the current state of the XBackend.
+    // +optional
+    // +listType=map
+    // +listMapKey=type
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+    // DiscoveredTools lists the tools discovered on this backend via
+    // MCP tools/list. Only tools with valid schemas are included.
+    // +optional
+    // +listType=map
+    // +listMapKey=name
+    DiscoveredTools []DiscoveredTool `json:"discoveredTools,omitempty"`
+
+    // LastDiscoveryTime is the timestamp of the last successful
+    // tools/list call to this backend.
+    // +optional
+    LastDiscoveryTime *metav1.Time `json:"lastDiscoveryTime,omitempty"`
+}
+
+type DiscoveredTool struct {
+    // Name is the unique identifier for the tool, as returned by
+    // the MCP server's tools/list response.
+    // +required
+    Name string `json:"name"`
+
+    // Title is the human-readable display name for the tool.
+    // Corresponds to the "title" field in MCP 2025-11-25.
+    // +optional
+    Title string `json:"title,omitempty"`
+
+    // Description is a human-readable description of the tool's
+    // functionality.
+    // +optional
+    Description string `json:"description,omitempty"`
+}
+```
+
+### XBackend Status Conditions
+
+The discovery controller sets the following conditions on `XBackend`:
+
+| Type | Status | Reason | Meaning |
+|------|--------|--------|---------|
+| `ToolsDiscovered` | `True` | `DiscoverySucceeded` | tools/list succeeded, all tools valid |
+| `ToolsDiscovered` | `True` | `PartiallyValid` | tools/list succeeded, but some tools had invalid schemas and were excluded |
+| `ToolsDiscovered` | `False` | `DiscoveryFailed` | tools/list call itself failed (backend unreachable, protocol error) |
+
+### Complete Example
+
+```yaml
+apiVersion: agentic.prototype.x-k8s.io/v0alpha0
+kind: XBackend
+metadata:
+  name: weather-service
+  namespace: default
+  annotations:
+    agentic.prototype.x-k8s.io/enable-discovery: "true"
+spec:
+  mcp:
+    serviceName: weather-mcp
+    port: 8080
+    path: /mcp
+status:
+  conditions:
+    - type: ToolsDiscovered
+      status: "True"
+      reason: DiscoverySucceeded
+      message: "Discovered 2 tools"
+      lastTransitionTime: "2026-03-16T10:30:00Z"
+  discoveredTools:
+    - name: get_forecast
+      title: "Weather Forecast"
+      description: "Get weather forecast for a location"
+    - name: get_alerts
+      title: "Weather Alerts"
+      description: "Get active weather alerts for a region"
+  lastDiscoveryTime: "2026-03-16T10:30:00Z"
+```
+
+An `XAccessPolicy` referencing a non-existent tool receives a Warning:
+
+```yaml
+apiVersion: agentic.prototype.x-k8s.io/v0alpha0
+kind: XAccessPolicy
+metadata:
+  name: agent-weather-access
+  namespace: default
+spec:
+  targetRefs:
+    - group: agentic.prototype.x-k8s.io
+      kind: XBackend
+      name: weather-service
+  rules:
+    - name: allow-forecast
+      source:
+        type: ServiceAccount
+        serviceAccount:
+          name: my-agent
+          namespace: default
+      authorization:
+        type: InlineTools
+        tools:
+          - "get_forecast"
+          - "get_humidity"  # Not discovered on backend
+status:
+  ancestors:
+    - ancestorRef:
+        group: agentic.prototype.x-k8s.io
+        kind: XBackend
+        name: weather-service
+      controllerName: agentic.prototype.x-k8s.io/controller
+      conditions:
+        - type: Accepted
+          status: "True"
+          reason: "Accepted"
+          message: "Policy accepted; unverified tools: get_humidity (not found in discovered tools for XBackend 'weather-service')"
+          lastTransitionTime: "2026-03-16T10:31:00Z"
+```
+
+## Implementation
+
+### Discovery Controller
+
+The discovery logic lives in `pkg/discovery/` within the existing
+`agentic-net-controller` binary as a new reconciler. Discovery is opt-in:
+the controller only processes XBackend resources that carry the
+`agentic.prototype.x-k8s.io/enable-discovery: "true"` annotation. Backends
+without this annotation are ignored, incurring zero overhead. The discovery
+controller can also be disabled entirely via a controller flag
+(`--enable-discovery=false`).
+
+For opted-in backends, it:
+
+1. **Watches** XBackend resources with the discovery annotation.
+2. **Connects** to each backend's MCP endpoint using the `mcp-go` client
+   library (`github.com/mark3labs/mcp-go`).
+3. **Calls** `tools/list`, handling pagination via cursor for backends with
+   many tools.
+4. **Validates** each tool's `inputSchema` against JSON Schema draft 2020-12.
+   Tools with invalid schemas are rejected and logged as Warning events on the
+   XBackend resource.
+5. **Updates** `XBackend.status.discoveredTools[]` with validated tools only.
+6. **Listens** for `tools/list_changed` notifications on backends that declare
+   the `listChanged` capability, triggering immediate re-discovery.
+7. **Falls back to polling** (configurable interval, default 30s) for backends
+   that don't support `list_changed`, and as a periodic consistency check.
+8. **Detects drift** by diffing new vs. old tool sets and emits Kubernetes
+   Events when tools are added, removed, or fail validation.
+
+### AccessPolicy Validation
+
+The existing AccessPolicy reconciler is extended to cross-reference
+`rules[].authorization.tools[]` against `XBackend.status.discoveredTools`:
+
+- If a tool name doesn't match, surface the unverified tool names in the
+  `Accepted` condition's `message` field on the relevant ancestor entry
+  (not rejection — the tool may not be discovered yet). This follows
+  Gateway API convention of using standard condition types rather than
+  introducing domain-specific ones.
+- Validation runs on both AccessPolicy changes and XBackend status changes,
+  so stale policies are detected when a backend removes a tool.
+
+### Phasing
+
+**Phase 1: In-cluster discovery (plain HTTP).** XBackend status extension,
+discovery controller, schema validation, `list_changed` support, drift
+detection. The controller connects to backends using the service name and port
+from `XBackend.spec.mcp` over plain HTTP. This is sufficient for in-cluster
+backends where trust is assumed. In mesh environments (e.g., Istio sidecar on
+the controller pod), mTLS is handled transparently at the infrastructure layer,
+so the plain HTTP implementation covers that case without extra work.
+
+Backends that require authentication for `tools/list` will fail gracefully:
+the XBackend receives a `ToolsDiscovered: False` condition with reason
+`DiscoveryFailed`, and AccessPolicy enforcement continues unaffected. Discovery
+is advisory — it does not gate the data plane. The operator sees an early
+signal that the backend is unreachable to the control plane, but runtime tool
+calls proceed normally as long as the agent's own credentials are valid.
+
+**Phase 2: External backends (TLS and authentication).** Extends the discovery
+controller to connect to backends outside the cluster, requiring TLS and
+potentially a credential reference for `tools/list` calls. The exact mechanism
+(e.g., a `credentialRef` on XBackend) is deferred pending stabilization of the
+Backend spec between KAN and the AI Gateway WG.
+
+**Phase 3: Policy validation.** AccessPolicy cross-referencing against
+discovered tools. When an XBackend has `ToolsDiscovered: False` (discovery
+failed or not yet attempted), any AccessPolicy targeting that backend includes
+a note in the `Accepted` condition message indicating which target backends
+had no discovery data available (e.g., "tools not verified against XBackend
+'weather-service': discovery unavailable"). This ensures operators have
+visibility without blocking policy acceptance.
+
+### Alternative: Separate Discovery Controller
+
+The discovery logic is encapsulated in `pkg/discovery/`, making extraction into
+a standalone `agentic-net-discovery` binary straightforward if the community
+prefers separation of concerns. This pattern, a dedicated controller whose
+sole job is to periodically fetch remote capabilities and cache them in CRD
+status, is used by the kagenti-operator's AgentCard controller, which fetches
+agent capability cards from `/.well-known/agent.json` endpoints and writes them
+to `AgentCard.status`. The benefit is fault isolation: a bug or crash in the
+discovery loop doesn't affect the main proxy management controller. The
+embedded approach is recommended to start because it avoids an additional
+deployment and simplifies RBAC.
+
+## Prior Art
+
+### Kuadrant/mcp-gateway MCPManager
+
+The MCP Gateway's broker connects to upstream MCP servers and calls
+`tools/list` on a reconciliation loop, maintaining an in-memory registry. It
+supports `tools/list_changed` for reactive re-discovery ([PR #329](https://github.com/Kuadrant/mcp-gateway/pull/329)) and prefixes
+tool names for federation. Its `MCPServerRegistration` CRD status stores only a
+tool count — sufficient for the gateway's needs but too minimal for KAN's
+policy validation use case.
+
+- *Pro:* Production-tested discovery loop with list_changed support.
+- *Pro:* Schema validation gap ([issue #662](https://github.com/Kuadrant/mcp-gateway/issues/662)) validates our design choice to
+  validate at the discovery layer.
+- *Con:* CRD status (count only) is insufficient for policy validation.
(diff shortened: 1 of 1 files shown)


docs/proposals/XXXX-ToolDiscovery.md
```diff
@@ -0,0 +1,441 @@
...
+connects to MCP backends, calls `tools/list`, validates the results, and
+surfaces discovered tools in `XBackend.status`. This enables policy validation
+against real backend state, operator visibility via `kubectl`, and a foundation
+for future capabilities like response filtering. Discovery is advisory (i.e.,
+it enhances observability but does not affect policy enforcement). Backends
+without discovery enabled continue to work exactly as they do today.
+
+## Non-Goals
+
+- **`tools/list` response filtering.** Filtering the `tools/list` response
```
david-martin, line 39: Good to call this out.
I wonder, should we be more explicit to avoid confusion here as there are 2 identities at play here.
The identity to use when this new controller calls tools/list for populating the backend status,
and the identity when an agent/user is calling tool list?
The latter tool/list filtering I see as dynamic at list time, performing an authZ function, and driven by the rules in AccessPolicy

docs/proposals/XXXX-ToolDiscovery.md
```diff
@@ -0,0 +1,441 @@
...
+  introducing domain-specific ones.
+- Validation runs on both AccessPolicy changes and XBackend status changes,
+  so stale policies are detected when a backend removes a tool.
+
+### Phasing
+
+**Phase 1: In-cluster discovery (plain HTTP).** XBackend status extension,
+discovery controller, schema validation, `list_changed` support, drift
+detection. The controller connects to backends using the service name and port
+from `XBackend.spec.mcp` over plain HTTP. This is sufficient for in-cluster
```
david-martin: https?

docs/proposals/XXXX-ToolDiscovery.md
```diff
@@ -0,0 +1,441 @@
...
+   library (`github.com/mark3labs/mcp-go`).
+3. **Calls** `tools/list`, handling pagination via cursor for backends with
+   many tools.
+4. **Validates** each tool's `inputSchema` against JSON Schema draft 2020-12.
+   Tools with invalid schemas are rejected and logged as Warning events on the
+   XBackend resource.
+5. **Updates** `XBackend.status.discoveredTools[]` with validated tools only.
+6. **Listens** for `tools/list_changed` notifications on backends that declare
+   the `listChanged` capability, triggering immediate re-discovery.
+7. **Falls back to polling** (configurable interval, default 30s) for backends
```
david-martin: 30s probably too aggresive. 
5 minutes maybe?

david-martin (commented):
Given the XBackend resource won't be in v1alpha1, we should hold this for now.
I see the value in having a controller being able to list and report back tools from mcp servers for more than AccessPolicy tool validation. For example, it can help with configuration validation, or surfacing of tools for informational purposes beyond auth.

I tend to like the standalone (extension?) approach, as making it a required piece could be a barrier to some agentic networking implementations, and could potentially be unwanted in some installations for various reasons (unwanted load, difficult to configure auth for some mcp servers)

/hold

docs/proposals/XXXX-ToolDiscovery.md
```diff
@@ -0,0 +1,441 @@
...
+    - name: get_alerts
+      title: "Weather Alerts"
+      description: "Get active weather alerts for a region"
+  lastDiscoveryTime: "2026-03-16T10:30:00Z"
+```
+
+An `XAccessPolicy` referencing a non-existent tool receives a Warning:
+
+```yaml
+apiVersion: agentic.prototype.x-k8s.io/v0alpha0
```
david-martin: agentic.networking.x-k8s.io

docs/proposals/XXXX-ToolDiscovery.md
```diff
@@ -0,0 +1,441 @@
...
+   the `listChanged` capability, triggering immediate re-discovery.
+7. **Falls back to polling** (configurable interval, default 30s) for backends
+   that don't support `list_changed`, and as a periodic consistency check.
+8. **Detects drift** by diffing new vs. old tool sets and emits Kubernetes
+   Events when tools are added, removed, or fail validation.
+
+### AccessPolicy Validation
+
+The existing AccessPolicy reconciler is extended to cross-reference
+`rules[].authorization.tools[]` against `XBackend.status.discoveredTools`:
```
david-martin: note the change to method based matching in v1alpha1 https://github.com/kubernetes-sigs/kube-agentic-networking/blob/d2507a72a61bdb00d01b3a0cf098478a648c24fc/api/v1alpha1/accesspolicy_types.go#L197
Will need to consider that and update here.

docs/proposals/XXXX-ToolDiscovery.md
rubambiza: Good catch. The distinction is important. With the direction we're considering based on community feedback, the discovery controller's identity is now fully decoupled: it has its own service account credentials for calling `tools/list` at reconcile time (a control-plane operation), while runtime `tools/list` filtering remains a data-plane concern driven by XAccessPolicy rules and the agent's identity. The latest revision makes this explicit in the Implementation section.

docs/proposals/XXXX-ToolDiscovery.md
rubambiza: Yes. Phase 1 assumes in-cluster where mTLS is handled transparently by mesh infrastructure (e.g., Istio sidecar). Phase 2 would add explicit TLS for external backends. Clarified in the updated Phasing section.

docs/proposals/XXXX-ToolDiscovery.md
rubambiza: Agreed. Updated to 5 minutes as the default, configurable via `spec.pollInterval` on the new XToolInventory resource.

docs/proposals/XXXX-ToolDiscovery.md
rubambiza: Fixed in https://github.com/kubernetes-sigs/kube-agentic-networking/pull/227/commits/b48bcbc2db82b9fb8e4b1562aaa2e3a46b0cda90. All examples and references now use `agentic.networking.x-k8s.io/v1alpha1` throughout.

docs/proposals/XXXX-ToolDiscovery.md
rubambiza: Indeed. The AccessPolicy validation section was written against the old `InlineTools` / `tools[]` model. Updated to align with v1alpha1 method-based matching: the cross-reference now validates `params[]` entries (for `tools/call` method rules) against `XToolInventory.status.discoveredTools[].name`.

docs/proposals/XXXX-ToolDiscovery.md
```diff
@@ -0,0 +1,542 @@
...
+| Date | Change |
+|------|--------|
+| 2026-05-25 | Major revision: replaced `XBackend.status.discoveredTools` with a dedicated `XToolInventory` CRD (1:1 with XBackend via `backendRef`). Updated all examples to `agentic.networking.x-k8s.io/v1alpha1` with method-based matching (`mcp.methods[].params[]`). Changed default poll interval from 30s to 5m. Added hybrid creation model (explicit + annotation-triggered). Motivated by community feedback from @keithmattix and @david-martin. |
+| 2026-04-06 | Added A2A scoping non-goal. |
+| 2026-03-16 | Initial proposal. |
+
+## Summary
+
+KAN has no mechanism to discover what tools an MCP backend exposes. Operators
+must manually enumerate tool names in `XAccessPolicy` rules, and there is no
```
LiorLieberman: would this still be true with CEL support?

docs/proposals/XXXX-ToolDiscovery.md
```diff
@@ -0,0 +1,542 @@
...
+feedback loop when tools are added, removed, or misconfigured on backends.
```
LiorLieberman: specifically on misconfigured - the usual approach for things like that is to communicate a status. 

Should we name this as an option and discuss pros/cons?

docs/proposals/XXXX-ToolDiscovery.md
```diff
@@ -0,0 +1,542 @@
...
+
+This proposal adds optional, opt-in tool discovery to KAN: a controller that
+connects to MCP backends, calls `tools/list`, validates the results, and
+surfaces discovered tools in a dedicated `XToolInventory` resource (one per
```
LiorLieberman, line 31: >dedicated `XToolInventory` resource (one per backend

How do we reason about this vs enhancing backend status?

docs/proposals/XXXX-ToolDiscovery.md
```diff
@@ -0,0 +1,542 @@
...
+    Hostname    *string `json:"hostname,omitempty"`
+    Port        int32   `json:"port"`
+    Path        string  `json:"path,omitempty"`
+}
+// No discoveredTools, no tool metadata.
+```
+
+This creates three concrete problems:
+
+**1. Stale access policies.** `XAccessPolicy` rules require operators to
```
LiorLieberman, line 71: technically, when you have an ALLOW AccessPolicy on a Backend, the rest that you dont allow is implicitly denied.  This leaves us with the following scenarios; 

1. A new tool is added on the backend -- it is denied automatically unless allowed explicitly -- which is the desired behavior (IMO)

2. Tool is removed from the backend -- then access policy is "stale". 

1 + 2 are relevant for rename as well.

Is the intent here for easier cleanup of stale policies?

LiorLieberman (commented):
First of all, thank you @rubambiza for a very detailed and informative proposal! It was easy to read. 

I left a few comments upon an initial full pass, however, I have more fundamental high level thoughts -- 

I agree that aggregating tools on backends async is useful. What I am not sure is how much value it gives to the user in terms of user facing API. 

Most of what I read here can be done as an implementation detail of a separate routine/process on the binary that does that discovery and validation instead of the gateway broadcasting tools/list when needed. I think thats what MCP-gateway (broker component IIRC?) is doing, but could be wrong

/cc @david-martin for additional thoughts/comments

rubambiza:
@LiorLieberman Re: value as a user facing API, beyond what's listed as value for operators, I think your suggestion in the last comment about cleaning up stale policies is also an additional value prop. You mentioned

>"Most of what I read here can be done as an implementation detail of a separate routine/process on the binary that does that discovery and validation instead of the gateway broadcasting tools/list when needed."

I am aligned on this. That would be an implementation detail, and you're right that it's what the MCP Gateway broker does today (in-memory caches in `internal/broker/upstream/manager.go`, including `tools/list_changed` handling).

The thing is, without a dedicated CRD/controller that is more programmable, it's unclear which backends would be involved in the discovery/validation process. For instance, does the routine contact all the backends, and if not, which ones? Worth noting that the MCP Gateway is actually moving discovery status out of its `MCPServerRegistration` CRD (kuadrant/mcp-gateway#1187) because broker-polled tool counts kept going stale, so runtime status now lives in the broker's `/status` endpoint. To me that reinforces the point: an internal routine on its own doesn't give producers (backends that opt-in) and consumers (operators using the info) a clear contract. The CRD provides that structure. See also the discussion on `#Alternative: Separate Discovery Controller Binary`.

docs/proposals/XXXX-ToolDiscovery.md
rubambiza: Good question, it's a bit more nuanced with CEL. Inline rules enumerate tool names explicitly in `mcp.methods[].params[]`, so they're directly cross-referenceable against discovered tools, and that's where staleness detection applies cleanly. CEL rules can match dynamically (the codebase has examples like `request.mcp.tool_name.startsWith('verify_')`), so a pattern-based rule wouldn't go stale the same way and isn't meaningfully validated against a discovered tool list. So the manual enumeration framing is really about Inline rules, and I've scoped the policy validation section (Phase 3) accordingly. Separately, @david-martin mentioned in a recent discussion that this CRD could be something the MCP Gateway leverages, since it's an "upstreaming" of similar/complementary functionality.

docs/proposals/XXXX-ToolDiscovery.md
rubambiza: I think `### Discovery data in XBackend.status` under the `## Alternatives considered` subsection discusses the pros/cons (mostly cons) based on community feedback and practical experiences. Moreover, the API _does_ communicate some status for misconfigurations (see the unverified tools example in the Accepted condition message). However, open to feedback if I may have misunderstood your question on naming it as an option and enumerating pros/cons.

docs/proposals/XXXX-ToolDiscovery.md
rubambiza: Same answer here as discussed in `### Discovery data in XBackend.status` under `## Alternatives considered`. To recap, we don't want to force every backend to opt-in and XBackend is in flux.

docs/proposals/XXXX-ToolDiscovery.md
rubambiza: Yes, stale policy cleanup would be much easier to do. I totally agree that "when you have an ALLOW AccessPolicy on a Backend, the rest that you dont allow is implicitly denied." Moreover, this proposal allows operators to know _what_ to ALLOW in the first place and how outdated/renamed/moved/removed it may be.

docs/proposals/XXXX-ToolDiscovery.md
```diff
@@ -0,0 +1,552 @@
...
+
+**CUJ 2: Operator writes a valid access policy.**
+The platform engineer writes an `XAccessPolicy` referencing tool names in
+`mcp.methods[].params[]`. The controller validates the tool names against
+the corresponding `XToolInventory.status.discoveredTools` and sets a Warning
+in the policy's `Accepted` condition message if any tool name doesn't match.
+
+**CUJ 3: Backend adds a new tool.**
+A tool developer deploys a new version of their MCP server with an additional
+tool. The discovery controller detects the change (via `tools/list_changed`
```
david-martin, line 119: What do you think about aligning with upcoming mcp spec changes, and using ttlm field instead of notifications? https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/

docs/proposals/XXXX-ToolDiscovery.md
```diff
@@ -0,0 +1,552 @@
...
+XBackend to trigger auto-creation. Within the poll interval,
+`kubectl get xtoolinventory weather-service-tools -o yaml` shows
+`status.discoveredTools` populated with the backend's tools. No manual
+enumeration required.
+
+**CUJ 2: Operator writes a valid access policy.**
+The platform engineer writes an `XAccessPolicy` referencing tool names in
+`mcp.methods[].params[]`. The controller validates the tool names against
+the corresponding `XToolInventory.status.discoveredTools` and sets a Warning
+in the policy's `Accepted` condition message if any tool name doesn't match.
```
shachartal: Seems that you are proposing that two different controllers modify the status subobject for a single CRD, which is commonly regarded as an anti-pattern. Or do we expect this to be a single controller managing both CRDs?

docs/proposals/XXXX-ToolDiscovery.md
```diff
@@ -0,0 +1,552 @@
...
+discovery enabled continue to work exactly as they do today.
+
+## Non-Goals
+
+- **`tools/list` response filtering.** Filtering the `tools/list` response
+  based on caller identity (so agents only see tools they're authorized to call)
+  is a natural next step but would require a separate proposal.
+- **Tool prefixing / federation.** Disambiguating tools with the same name
+  across multiple backends (e.g., via prefix namespacing) is out of scope.
+- **Full tool schema storage.** Storing `inputSchema` and `outputSchema` in
```
shachartal, line 44: I agree with this being out of scope for this proposal. It would be good to find another mean to supply said schemas (doesn't have to be in the first phase of course) - e.g. serving the schemas via an Extension API Server.

docs/proposals/XXXX-ToolDiscovery.md
rubambiza: I looked over the SEP and the release candidate, and I agree on alignment, with two caveats. Like the proposed change, I think it should be an AND rather than relying on `ttlMs` alone. I understand that leaning primarily on `ttlMs` would cut a lot of the refetch traffic, and I'm open to using it as the primary mechanism as long as we make it clear to implementers that we as a community are okay with that trade-off for what `XToolInventory` is meant to achieve.

Secondly, and related, I'm wary of relying on `ttlMs` alone because of the Security Implications section, which notes that a misconfigured or malicious server could set an excessively long TTL. That would defeat most of the purpose of the discovery controller in giving policy authors fresh data. So I think the answer is somewhere in the middle: honor `ttlMs` as the primary mechanism within reason, while also leveraging notifications for servers that provide them.

I've pushed a revision that reflects this: the controller refetches after `min(ttlMs, spec.maxPollInterval)`, treats `tools/list_changed` as immediate invalidation, and falls back to `spec.pollInterval` when a backend advertises neither. The `maxPollInterval` bound is what guards against the long-TTL case. I also added it as a consensus point since it's a trade-off worth an explicit nod from the group. What do you think?

docs/proposals/XXXX-ToolDiscovery.md
rubambiza: I think there's a small misunderstanding here. The two controllers don't both write to `XToolInventory.status`. The discovery controller owns and writes `XToolInventory.status`. The XAccessPolicy controller only *reads* `XToolInventory.status.discoveredTools` and then writes to its own `XAccessPolicy.status`. So each controller writes only to the status of the CRD it owns. No shared-writer anti-pattern, but I appreciate the callout as it would be.

docs/proposals/XXXX-ToolDiscovery.md
rubambiza: Can you kindly elaborate on "serving the schemas via an Extension API Server"? I'm trying to understand the value proposition and who we imagine the users to be.

shachartal: I would imagine the use cases can be anywhere between troubleshooting ,and additional tools in the ecosystem wanting to evaluate the same without having to implement the same flow to fetch the schemsa.

Value prop is focused around being able to consume the schema with ~ `kubectl get mcpinputschema ...`, through the API server on the management plane, rather than by interacting with the application over the network/data plane (ostensibly this would also require credentials and network policy allowing access). As an example, see OpenShift Image Streams, which exposes newly pushed image tags via aggregated API, and removes the need for CD tools to interact with, or even have network access to image registries.

david-martin:
I think there is value in a few places here for different personas, but there's also a big caveat that
should be allowed for.

First, the platform engineer. By deploying the controller, they can get a central view of all tools
across all mcp backends. Useful for visibility and troubleshooting to have that information available.

Then for the user defining an AccessPolicy. If the controller is highlighting drift in the actual tools
available vs. what's in AccessPolicy rules, this can help with catching potential security regressions.

If there is general agreement on these value points, I think the proposal could collapse the phased
approach so that visibility and drift detection (including the AccessPolicy cross-referencing currently
in Phase 3) land together, and get feedback quicker.

Now, the caveat. Tools can be dynamic. There can be specific tools available depending on the user. This
is called out clearly in the idea of MCP server cards SEP at https://github.com/modelcontextprotocol/modelcontextprotocol/blob/sep/mcp-server-cards/seps/2127-mcp-server-cards.md.
This wording is relevant:
```
Primitive exclusion as a safety property: Cards deliberately exclude primitive definitions (tools,
resources, prompts) so clients cannot trust a static manifest for access-control or safety decisions;
primitives are always validated at runtime via standard list operations.
```

There's an issue that discusses this further at
https://github.com/modelcontextprotocol/modelcontextprotocol/issues/540

The proposal should call this out as a known limitation in my opinion (make it clear it's something
that's actively discussed in the mcp community).

LiorLieberman:
I took another read, my overall take here is that I still am not sure about the value of adding APIs or the viability of the currently described CUJs. 

Before sharing more on my reasoning, I do think that some kind of a cli extension to kubectl or gwctl can help achieve some of the motivation here, AND that implementing some kind of offline caching of tools can help with other things like tools/list requests but this requires only implementation, no api change

Trying to unpack the goal from the summary -- 

> KAN has no mechanism to discover what tools an MCP backend exposes. Operators
who write `Inline` `XAccessPolicy` rules must manually enumerate tool names,
and there is no feedback loop when tools are added, removed, or misconfigured
on backends.

What I read here is that (a) KAN does not track what mcp tools are exposed on backends. (b) operators author authorization rules and enumerate them manually and (c) nothing communicates changes of tools exposed on backends.

For (a) - if we needed to track what tools are exposed on each backend - thats an implementation work, not api. If we need a knob to enable it - that could be a flag, I am not sure a CRD is needed. And we would need to call out why we need it - for example, one useful thing is to "respond fast to tools/list command on the gateway instead of broadcast tools/list to all backend behind it"

For (b) I think that is by design. CEL is useful to make it less manual and more dynamic. AuthZ rules should be configured separately than what backends are exposed, so I see the no-link between tools (or just REST endpoints in general) exposed to the AuthZ rules on them as WAI.

For (c) - I am also not convinced this is necessary to communicate tools has changed on the backend. I want to again come back to just REST endpoints - we dont have anywhere something that is communicating what REST endpoints a backend/service expose. And I am not aware of prior art that tries to automate or link authorization polices to it. Backends can change all the time, and tools exposed are just another code change. In a security-aware org, you will have rules that provides zero-trust OR just allow lists, like the ALLOW list we have with AccessPolicy.

The as it is right now, advocates for adding some validation on accessPolicy whether the tools indeed exist on the backend, which I believe is not the behavior we want for the reason I shared above. And we certainly do not want to automate "cleanup" or "removal" or accesspolicy rules in case they are not exposed, cause we have no way to know the intent of the user,  what if they want the rules to be there, regardless and just in case a backend decides to expose them.

Regarding the discoverability part here - meaning how would an author of an AccessPolicy know what tools are exposed - I dont think its reasonable to assume a CUJ in which a policy author is deploying another CRD (tool inventory), and wait for reconciliation just to know what tools they have. More realistically, I think the policy author is either (a) the tool owner, in this case they know the tools or (b) can just read docs (like we all read open-api or rest docs) or issue a tools/list request to find out.

rubambiza:
Thanks Dave, good points on both fronts.

On collapsing the phases: agree on the value points for both personas (central visibility for the platform engineer, drift-as-security-signal for the policy author), and I'm on board. To be precise about what moves: Phase 1 already includes discovery, `XToolInventory` status, and drift detection. What's currently Phase 3 is the XAccessPolicy cross-referencing (surfacing unverified tool names in the policy's `Accepted` condition). I've folded that into the first deliverable so visibility, drift detection, and policy cross-referencing land together. The one thing I kept separate is Phase 2 (external backends needing TLS/auth).

On that Phase 2 point, it's worth noting the Backend resource has since landed as Implementable and shipped experimentally in Gateway API v1.6.0 (GEP-4488), and MCP is now a named `BackendProtocol` value. So the resource itself is no longer the open question. What's still genuinely deferred is narrower: the MCP-specific protocol options and the credential mechanism for authenticated `tools/list` calls to external backends, which the GEP marks as TODO.

Incidentally, GEP-4488's status model reinforces the separate-CRD decision. Backend's status is `BackendParentStatus`, keyed per controller and parent association (routing-oriented), and the GEP deliberately keeps primitive/tool concerns off the resource. That's consistent with keeping discovered tools in a dedicated `XToolInventory` rather than on the backend.

On MCP Server Cards: good pointer. I read through SEP-2127 and issue #540, and I think the relationship is worth spelling out, but as a complement rather than an overlap. Server Cards are a static document served without a connection, answering "where and how do I connect" (transports, protocol versions), and they deliberately exclude primitives. XToolInventory is a cache of a live `tools/list` result obtained *with* a connection, by the controller's single control-plane identity, to support policy authoring. So we sit on the same side of the line the SEP draws: "primitives are always validated at runtime via standard list operations." We're not trying to make primitives statically advertisable, which is the thing the SEP defers. If anything, I'd frame SEP-2127 as corroborating our design constraint, since it independently lands on the same conclusion that motivates our advisory-only posture.

On the caveat itself, you're right, and I've called it out explicitly as a known limitation. Because the controller discovers with one identity, it captures a single "control-plane variant" of the tool list. Tools gated on the caller's identity or session won't necessarily match what a specific user sees at runtime, so the drift signal is advisory and can have false positives. I've noted that this is an actively discussed topic in the MCP community (SEP-2127's primitive exclusion and #540) and that it's the reason we never gate the data plane on discovery.

All of this is in the latest revision (commits 6865240 and 7bffe9c): the phase collapse, the "Relationship to MCP Server Cards (SEP-2127)" and "Known Limitation: dynamic and per-caller tools" sections, and the updated Phase 2 wording reflecting GEP-4488.

rubambiza:
Thanks for the thorough read, Lior. Let me take the summary and each point in turn.

> I do think that some kind of a cli extension to kubectl or gwctl can help achieve some of the motivation here, AND that implementing some kind of offline caching of tools can help with other things like tools/list requests but this requires only implementation, no api change

I agree these look largely like implementation concerns on the surface, but I'm keeping the bigger picture in mind. If we focus only on implementation, a few things stay undefined, and they're the same questions worth asking: how often does the routine loop? When do we consider results stale? What's the contract if we later want to support something like the schema-storage extension Shachar advocated for? Those are contract questions, not just code. The CRD gives us a place to answer them explicitly (`spec.pollInterval`, `spec.maxPollInterval` / `status.observedTTL` for staleness, and a status shape that can grow). I see it as a starting contract we can scrutinize as a POC in action, the same way the KAN reference implementation let us pressure-test other decisions. A kubectl/gwctl extension is a nice complement for read access, but it's a read tool; it doesn't by itself define loop cadence, staleness, or an extension point other controllers can consume.

> For (a) ... one useful thing is to "respond fast to tools/list command on the gateway instead of broadcast tools/list to all backend behind it"

Good questions, and they get at why I think there's a contract worth defining rather than pure implementation. On the "respond fast at the gateway instead of broadcasting to all backends" idea, I think that's genuinely useful. The natural follow-up is where the gateway reads that cached information from. If it's a side database, then every KAN implementer would need to stand one up and operate it. Leaning on a CRD lets us reuse infrastructure that's already there (etcd, the API server, informers), so it's closer to seamless for implementers, and it gives the "who enables it / what's the trust boundary" questions a concrete home (the opt-in annotation and the advisory-only status).

> For (b) ... I see the no-link between tools ... exposed to the AuthZ rules on them as WAI.

("WAI" = working as intended, for anyone like me who did not know the acronym until today and following along :D) I concur CEL makes authorization less manual and more dynamic. That's exactly why the July 2nd revision scoped validation to `Inline` rules, where the rule definition *is* tied to knowing which tools are available. Concretely, in our own quickstart the Inline policy enumerates tool names the author had to know up front:

```yaml
authorization:
  type: Inline
  mcp:
    methods:
      - name: tools/call
        params:
          - "get-sum"
          - "get-tiny-image"
```

(from [`site-src/guides/quickstart/policy/e2e.yaml`](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/main/site-src/guides/quickstart/policy/e2e.yaml)). For CEL, I fully agree the decoupling is WAI, and we don't touch it.

> For (c) ... we dont have anywhere something that is communicating what REST endpoints a backend/service expose ...

I'd gently push on the REST comparison. We do have ways to know which services are deployed and which endpoints to curl. But I think REST endpoints are both too granular and too broad an abstraction for this comparison. MCP forces us to write AuthZ rules against *tools*, not backends nor their REST paths. In fact, MCP typically multiplexes many tools behind a single endpoint (e.g. `/mcp`), so the backend/route abstraction Gateway API already gives us can't express tool-level intent on its own. This CRD cuts through both to get at that level, at least for Inline rules.

> And we certainly do not want to automate "cleanup" or "removal" or accesspolicy rules ...

Fully onboard, no automated cleanup or removal. Too many unknowns about intent; the rules may be there deliberately, waiting for a backend to (re)expose a tool. So let's forego that entirely and scope the value to what you and Dave converged on: visibility and troubleshooting for the platform engineer, and drift highlighting for the policy author, without acting on it. That's captured in the latest revision, along with the important caveat Dave raised about dynamic/per-caller tools, which I've written up as a [Known Limitation](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/c94e01e/docs/proposals/XXXX-ToolDiscovery.md#known-limitation-dynamic-and-per-caller-tools) section.

> I dont think its reasonable to assume a CUJ in which a policy author is deploying another CRD (tool inventory), and wait for reconciliation just to know what tools they have.

Agreed, and I don't see the policy author as necessarily the same person deploying the CRD. That's exactly why we argue for the opt-in hybrid model: the operator/platform engineer creates the XToolInventory (or the controller auto-creates it via annotation), and the policy author is a separate consumer of the resulting visibility. I've tightened CUJ 2 (commit c94e01e) so the persona split is explicit and to note the policy author never has to deploy or reconcile a discovery resource themselves.

> I think the policy author is either (a) the tool owner, in this case they know the tools or (b) can just read docs ... or issue a tools/list request to find out

Funny enough, I've been on both sides of this. At a startup I was the cluster owner, namespace owner, and workload owner all at once, and I had 7-8 Gateway API tabs open trying to figure out how to reach one of our app endpoints on GKE. But I've also worked on clusters at orgs like IBM where the cluster owner could not care less which tools are deployed under each namespace. So, perhaps some of my adamance about this CRD is that I've felt the pain from both roles.

david-martin:
> GEP-4488

https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-4894/index.md ?

david-martin:
@rubambiza What the unique value here boils down to is:

- if you use inline tools in your access policy
- and you don't have user specific tool lists from an mcp server

you can get a hint on the accesspolicy status that a tool is no longer available at the mcp server, or is perhaps mispelled/renamed.

There are other pieces of value, but they are not necessarily unique, and could be solved in other ways (already alluded to in Liors comments e.g. a cli for listing tools)

2 thoughts:

- I wonder if this justifies a new CRD
- Is that unique value of sufficient value that it will have users (and sufficient interest to be maintained as part of this project)

LiorLieberman:
@rubambiza Based on discussions in the community meeting on 13th Aug - are you planning to split this to immediate deliverables + issues to follow up as discussed?

rubambiza:
@LiorLieberman Yes, I am working on creating the associated issues, hoping to have them ready after the upcoming US holiday.

rubambiza:
@LiorLieberman Following up now that the holiday's behind us. Here's how I'm splitting it. As discussed on Aug 13 (and per Dave's question on whether the unique value justified a new CRD), I'm dropping the dedicated `XToolInventory` CRD and delivering the value without a new API:

**Immediate:**

1. **A CLI extension to list the tools a backend exposes.** Additive to the existing `agentic-net-tool` binary, giving platform engineers the visibility/troubleshooting value directly, no CRD, no reconciliation. The one dependency is an MCP client to issue `tools/list` (we parse MCP via the Envoy filter today, not in Go), and `XBackend` already tells us which backends exist, so there's no separate registry to build.

2. **Gateway-side caching of `tools/list`** so the gateway can answer instead of broadcasting to every backend. I want to file this as a feasibility investigation rather than committed implementation, because the "answer it at the gateway" part is more involved than it first looks. Doing it in the Envoy MCP filter would be a core-Envoy change (upstream review, potentially months), so the path we'd actually own is a Go-side cache component, which means introducing MCP handling in Go. This is also where the `ttlMs` freshness discussion Dave and I had lands: whatever populates the cache honors `min(ttlMs, maxPoll)`, treats `tools/list_changed` as invalidation, and bounds the malicious/long-TTL case.

**Follow-up:** Shachar's suggestion of serving tool schemas via an Extension API Server. It was always independent of the CRD, so dropping the CRD doesn't affect it. I'll file it separately to keep the idea alive rather than fold it into the immediate work.

I also looked at Sandor's MCP-registry idea. I think it solves a different layer (which MCP *servers* exist) than what we converged on here (which *tools* a known backend exposes), and KAN already has that server-level view via `XBackend`, so I don't think we need to build a separate registry for this.

I'll link the issues back here as I open them.

rubambiza:
Filed the split as issues:

- #339 — CLI subcommand to list the tools an MCP backend exposes (immediate)
- #340 — Investigate gateway-side caching of `tools/list` responses (immediate)
- #341 — Follow-up: serve MCP tool schemas via an Extension API Server

cc @LiorLieberman @david-martin @shachartal

rubambiza:
Closing in favor of the deliverables listed in [above](https://github.com/kubernetes-sigs/kube-agentic-networking/pull/227#issuecomment-5620405056)

PR #228, opened 2026-04-07
Fix the issue where `make api-ref-docs` was generating an empty reference page

haiyanmeng:
**What type of PR is this?**

/kind bug

Modify `hack/mkdocs/generate.sh` to use the local api directory directly instead of copying it to a temporary directory outside the project root. This allows `crd-ref-docs` to correctly resolve Go packages.

Fixes #226

```release-note
NONE
```

### DIFF
--- a/hack/mkdocs/generate.sh
+++ b/hack/mkdocs/generate.sh
@@ -34,18 +34,7 @@ declare -a arr=(
     "main"
 )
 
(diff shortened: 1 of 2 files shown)


rubambiza (approved):
LGTM.

LiorLieberman (commented):
/approve
/lgtm

PR #231, opened 2026-04-08
feat: Add translation logic to support custom TLS and frontendValidation in Gateway

ziyue-101:
/kind feature

## Description
This PR introduces support for Gateway TLS and `frontendValidation` (mTLS) within the `kube-agentic-networking` translator. It enables the controller to dynamically resolve server certificates and CA bundles to configure secure client-to-gateway connections.

### Key Changes
- **Downstream TLS Support**: Implemented translation logic to convert Gateway `TLS` and `FrontendTLSValidation` fields into Envoy `DownstreamTlsContext` configurations.
- **mTLS Resolution**: Supports both `CertificateRefs` for server identities and `CACertificateRefs` for client certificate validation (supporting both default and per-port overrides).
- **Documentation**: Added a new guide in `site-src` explaining how to configure mTLS correctly.
- **Test Coverage**: 
    - New unit tests for TLS resolution in `pkg/translator`.
    - Refactored E2E tests using Go subtests to cover positive and negative mTLS scenarios.
- **Status Reporting**: The Gateway status now provides feedback if `CACertificateRefs` is missing during an intended mTLS setup.

## How Has This Been Tested?
- **Unit Tests**: `go test ./pkg/translator/...` passes with new TLS test cases.
- **E2E Tests**: `make test-e2e` (specifically `TestGatewayTLS`) verifies that the gateway correctly enforces mTLS and rejects untrusted clients.
- **Confromance Tests** : results in https://github.com/kubernetes-sigs/kube-agentic-networking/issues/94#issuecomment-4331532024

Ref #94

```release-note
None, the API field is not exposed to users yet.
```

### DIFF
--- a/pkg/translator/common_test.go
+++ b/pkg/translator/common_test.go
@@ -26,8 +26,11 @@ import (
 	agenticv0alpha0 "sigs.k8s.io/kube-agentic-networking/api/v0alpha0"
 )
 
-func newTestGateway(name, ns string) *gatewayv1.Gateway {
-	return &gatewayv1.Gateway{
+// Suppress linter as ns always receives the same value: "quickstart-ns"
+//
+//nolint:unparam
+func newTestGateway(name, ns string, certRefs []gatewayv1.SecretObjectReference, frontendConfig *gatewayv1.FrontendTLSConfig) *gatewayv1.Gateway {
+	gw := &gatewayv1.Gateway{
 		ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: ns},
 		Spec: gatewayv1.GatewaySpec{
 			GatewayClassName: "cloud-provider-kind",
@@ -41,8 +44,22 @@ func newTestGateway(name, ns string) *gatewayv1.Gateway {
 			}},
 		},
 	}
+	if len(certRefs) > 0 {
+		gw.Spec.Listeners[0].TLS = &gatewayv1.ListenerTLSConfig{
+			CertificateRefs: certRefs,
+		}
+	}
+	if frontendConfig != nil {
+		gw.Spec.TLS = &gatewayv1.GatewayTLSConfig{
+			Frontend: frontendConfig,
+		}
+	}
+	return gw
 }
 
+// Suppress linter as ns always receives the same value: "quickstart-ns"
+//
+//nolint:unparam
 func newTestBackend(name, ns string) *agenticv0alpha0.XBackend {
 	return &agenticv0alpha0.XBackend{
 		ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: ns},
@@ -123,6 +140,10 @@ func newTestAccessPolicy(name, ns, targetName, targetKind, principal string) *ag
 	}
 }
 
+// Suppress linter as ns always receives the same value: "quickstart-ns"
+// Suppress linter as port always receives 3001
+//
+//nolint:unparam
 func newTestService(name, ns string, port int32) *corev1.Service {
 	return &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: ns},
--- a/pkg/translator/listener.go
+++ b/pkg/translator/listener.go
@@ -19,6 +19,7 @@ package translator
 import (
 	"crypto/sha256"
 	"encoding/json"
+	"encoding/pem"
 	"fmt"
 	"time"
 
@@ -40,6 +41,7 @@ import (
 	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/structpb"
 	"google.golang.org/protobuf/types/known/wrapperspb"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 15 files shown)


haiyanmeng:
@ziyue-101 , can you add e2e test coverage in this PR?

ziyue-101:
> @ziyue-101 , can you add e2e test coverage in this PR?

Done

haiyanmeng (commented):
@ziyue-101 , please take a look at the presubmit failure.

ziyue-101:
> @ziyue-101 , please take a look at the presubmit failure.

Uploaded a new version. The test should pass this time.

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
 				})
 			}
 
+			// secret extraction
```
haiyanmeng, line 210: `buildEnvoyResourcesForGateway` is getting two big with this change. Can you wrap your change into a separate method, and call it in `buildEnvoyResourcesForGateway`?

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+								ObservedGeneration: gateway.Generation,
+							})
+							caValid = false
+						} else {
+							caStr, hasCA := cm.Data["ca.crt"]
+							if !hasCA {
+								meta.SetStatusCondition(&listenerStatus.Conditions, metav1.Condition{
+									Type:               string(gatewayv1.ListenerConditionResolvedRefs),
+									Status:             metav1.ConditionFalse,
+									Reason:             "InvalidCACertificateRef",
```
haiyanmeng: Let us use the constant `ListenerReasonInvalidCACertificateRef` defined in the Gateway API

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+					}
+					if !matchedPort && gateway.Spec.TLS.Frontend.Default.Validation != nil && len(gateway.Spec.TLS.Frontend.Default.Validation.CACertificateRefs) > 0 {
+						// TODO: should we support mutiple CAs?
+						caRef = &gateway.Spec.TLS.Frontend.Default.Validation.CACertificateRefs[0]
+					}
+				}
+
+				if certRef != nil && caRef == nil {
+					meta.SetStatusCondition(&listenerStatus.Conditions, metav1.Condition{
+						Type:               string(gatewayv1.ListenerConditionProgrammed),
```
haiyanmeng: Should the type be `string(gatewayv1.ListenerConditionResolvedRefs)`?

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+						Status:             metav1.ConditionFalse,
+						Reason:             "InvalidCACertificateRef",
```
haiyanmeng: Use `ListenerReasonInvalidCACertificateRef` from Gateway API

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+						ns = string(*caRef.Namespace)
+					}
+					configMapName := string(caRef.Name)
+					sdsName := fmt.Sprintf("%s-%s", ns, configMapName)
+
+					if caRef.Kind != "" && caRef.Kind != "ConfigMap" {
+						meta.SetStatusCondition(&listenerStatus.Conditions, metav1.Condition{
+							Type:               string(gatewayv1.ListenerConditionResolvedRefs),
+							Status:             metav1.ConditionFalse,
+							Reason:             "InvalidCACertificateKind",
```
haiyanmeng: Use `ListenerReasonInvalidCACertificateKind` defined in Gateway API

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+							ObservedGeneration: gateway.Generation,
+						})
+						caValid = false
+					} else {
+						cm, err := t.configMapLister.ConfigMaps(ns).Get(configMapName)
+						if err != nil {
+							meta.SetStatusCondition(&listenerStatus.Conditions, metav1.Condition{
+								Type:               string(gatewayv1.ListenerConditionResolvedRefs),
+								Status:             metav1.ConditionFalse,
+								Reason:             "InvalidCACertificateRef",
```
haiyanmeng: Let us use the constant ListenerReasonInvalidCACertificateRef defined in the Gateway API

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+						meta.SetStatusCondition(&listenerStatus.Conditions, metav1.Condition{
+							Type:               string(gatewayv1.ListenerConditionResolvedRefs),
+							Status:             metav1.ConditionFalse,
+							Reason:             string(gatewayv1.ListenerReasonInvalidCertificateRef),
+							Message:            fmt.Sprintf("Failed to get secret %s: %v", secretName, err),
+							ObservedGeneration: gateway.Generation,
+						})
+						certValid = false
+					} else {
+						certBytes, hasCert := secret.Data["tls.crt"]
```
haiyanmeng: Use https://github.com/kubernetes/kubernetes/blob/03779bbd00da25c7fcd03711ddfe466a3322e1d7/pkg/apis/core/types.go#L6707

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+						keyBytes, hasKey := secret.Data["tls.key"]
```
haiyanmeng: https://github.com/kubernetes/kubernetes/blob/03779bbd00da25c7fcd03711ddfe466a3322e1d7/pkg/apis/core/types.go#L6709

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+				}
+
+				certValid := true
+				caValid := true
+
+				if certRef != nil {
+					ns := gateway.Namespace
+					if certRef.Namespace != nil {
+						ns = string(*certRef.Namespace)
+					}
```
haiyanmeng: Cross-namespace references to Secrets must be explicitly allowed by a ReferenceGrant.

If a Gateway in namespace A references a Secret in namespace B, there must be a ReferenceGrant created in namespace B that allows:

- From: Gateway resources in namespace A.
- To: Secret resources in namespace B.

So we need to add a check here to ensure a valid ReferenceGrant exists when certRef.Namespace != gateway.Namespace.

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+											Specifier: &corev3.DataSource_InlineBytes{InlineBytes: keyBytes},
+										},
+									},
+								},
+							}
+							envoySecrets = append(envoySecrets, envoySecret)
+							seenSecrets[sdsName] = true
+						}
+					}
+				}
```
haiyanmeng: The nesting if-else block is difficult to read. Please refactor to reduce nesting and improve readability.

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+										},
+									},
+								}
+								envoySecrets = append(envoySecrets, envoySecret)
+								seenSecrets[sdsName] = true
+							}
+						}
+					}
+				}
+
```
haiyanmeng: The nesting if-else block is difficult to read. Please refactor to reduce nesting and improve readability.

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+							envoySecrets = append(envoySecrets, envoySecret)
+							seenSecrets[sdsName] = true
+						}
+					}
+				}
+
+				if caRef != nil {
+					ns := gateway.Namespace
+					if caRef.Namespace != nil {
+						ns = string(*caRef.Namespace)
```
haiyanmeng: We need check ReferenceGrant here too

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+							meta.SetStatusCondition(&listenerStatus.Conditions, metav1.Condition{
+								Type:               string(gatewayv1.ListenerConditionResolvedRefs),
+								Status:             metav1.ConditionFalse,
+								Reason:             "InvalidCACertificateRef",
+								Message:            fmt.Sprintf("Failed to get configmap %s: %v", configMapName, err),
+								ObservedGeneration: gateway.Generation,
+							})
+							caValid = false
+						} else {
+							caStr, hasCA := cm.Data["ca.crt"]
```
haiyanmeng: Use https://github.com/kubernetes/kubernetes/blob/03779bbd00da25c7fcd03711ddfe466a3322e1d7/pkg/apis/core/types.go#L6653

pkg/translator/listener.go
```diff
@@ -606,39 +606,106 @@ func buildRouterFilter() (*hcm.HttpFilter, error) {
...
+
+	// 1. Server Certificate
+	var serverSecretName string
+	if lis.TLS != nil && len(lis.TLS.CertificateRefs) > 0 {
+		// TODO: should we support mutiple certs?
+		ref := lis.TLS.CertificateRefs[0]
+		if ref.Namespace != nil {
+			serverSecretName = fmt.Sprintf("%s-%s", *ref.Namespace, ref.Name)
+		} else {
+			serverSecretName = fmt.Sprintf("%s-%s", gateway.Namespace, ref.Name)
```
haiyanmeng: The logic to construct the SDS secret name `fmt.Sprintf("%s-%s", namespace, name)` is duplicated in both pkg/translator/translator.go and pkg/translator/listener.go. To follow the DRY principle and prevent future bugs if the naming scheme changes, we should extract this into a helper function.

pkg/translator/listener.go
```diff
@@ -606,39 +606,106 @@ func buildRouterFilter() (*hcm.HttpFilter, error) {
...
+		}
+	}
+
+	if clientValidationRef != nil {
+		var clientSecretName string
+		if clientValidationRef.Namespace != nil {
+			clientSecretName = fmt.Sprintf("%s-%s", *clientValidationRef.Namespace, clientValidationRef.Name)
+		} else {
+			clientSecretName = fmt.Sprintf("%s-%s", gateway.Namespace, clientValidationRef.Name)
+		}
```
haiyanmeng: These logic are duplicate with the logic added in pkg/translator/translator.go. Can we look into how to address this?

tests/e2e/tls_helpers.go
```diff
@@ -0,0 +1,167 @@
...
+
+// createCACertConfigMap creates a ConfigMap containing the CA certificate.
+func createCACertConfigMap(kc kubernetes.Interface, namespace, name string, caCertPEM []byte) error {
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: namespace,
+			Name:      name,
+		},
+		Data: map[string]string{
+			"ca.crt": string(caCertPEM),
```
haiyanmeng: Use the constant defined in corev1

tests/e2e/controller_test.go
```diff
@@ -757,3 +788,125 @@ func (m *mcpTestSession) assertToolCall(t *testing.T, toolName, toolArgs string,
...
+	err = createCACertConfigMap(kc, namespace, "gateway-client-ca", caCertPEM1)
+	if err != nil {
+		t.Fatalf("failed to create CA configmap: %v", err)
+	}
+
+	// Apply gateway-tls.yaml to enable mTLS on the common gateway
+	applyToNamespace(t, "testdata/gateway-tls.yaml", namespace)
+
+	// Deploy tester pod
+	applyToNamespace(t, "testdata/tester-pod.yaml", namespace)
```
haiyanmeng: `tester-pod.yaml` already has certificates mounted (e.g., for SPIFFE identity), they might interfere with the test or curl might accidentally fall back to them if something goes wrong with the custom paths we are writing to.

We should create a dedicated test pod definition for this test case that does not mount any default certificates.

tests/e2e/controller_test.go
```diff
@@ -757,3 +788,125 @@ func (m *mcpTestSession) assertToolCall(t *testing.T, toolName, toolArgs string,
...
+
+	podName := "e2e-tester"
+
+	// Write files to pod
+	runKubectl(t, "exec", podName, "-n", namespace, "--", "mkdir", "-p", "/tmp/mtls-client")
+	writePodFile(t, namespace, podName, testClientCertPath, clientCert1)
+	writePodFile(t, namespace, podName, testClientKeyPath, clientKey1)
+	writePodFile(t, namespace, podName, testClientCAPath, caCertPEM1)
+
+	// Test case 1: Client cert signed by trusted CA (should succeed)
```
haiyanmeng: Can we extend the test case 1 to include a XAccessPolicy targeting the Gateway, and  some `mcp.assertToolCall` calls?

pkg/translator/listener.go
```diff
@@ -606,39 +606,106 @@ func buildRouterFilter() (*hcm.HttpFilter, error) {
...
+		RequireClientCertificate: wrapperspb.Bool(true),
+	}
+
+	commonTLS := tlsContext.GetCommonTlsContext()
+
+	// 1. Server Certificate
+	var serverSecretName string
+	if lis.TLS != nil && len(lis.TLS.CertificateRefs) > 0 {
+		// TODO: should we support mutiple certs?
+		ref := lis.TLS.CertificateRefs[0]
```
bowei: If we only support a single cert, it's better to signal an error that additional certs are ignored.

pkg/translator/listener.go
```diff
@@ -606,39 +606,106 @@ func buildRouterFilter() (*hcm.HttpFilter, error) {
...
+				SdsConfig: &corev3.ConfigSource{
+					ResourceApiVersion: corev3.ApiVersion_V3,
+					ConfigSourceSpecifier: &corev3.ConfigSource_Ads{
+						Ads: &corev3.AggregatedConfigSource{},
+					},
+				},
+			},
+		}
+	} else {
+		commonTLS.ValidationContextType = &tlsv3.CommonTlsContext_ValidationContextSdsSecretConfig{
```
bowei, line 834: this will be easier to read if you factored the struct construction as helper methods with the fill-ins as parameters:


func newValidationContextTypeForSecret(secretName string)

this comment applies to the other large nested struct definitions

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
 				})
 			}
 
+			// secret extraction
+			if listener.Protocol == gatewayv1.HTTPSProtocolType {
+				var certRef *gatewayv1.SecretObjectReference
+				var caRef *gatewayv1.ObjectReference
+
+				if listener.TLS != nil && len(listener.TLS.CertificateRefs) > 0 {
+					// TODO: should we support mutiple certs?
```
bowei: validate that we don't have > 1 cert and then we can make the TODO explicit

pkg/translator/translator.go
```diff
@@ -226,6 +233,162 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+								},
+							}
+							envoySecrets = append(envoySecrets, envoySecret)
+							seenSecrets[sdsName] = true
+						}
+					}
+				}
+
+				if caRef != nil {
+					ns := gateway.Namespace
```
bowei: this needs to be refactored into separate methods

if caRef != nil {
  err := loadCA(...)

}


func loadCA() { ... }

pkg/translator/listener.go
ziyue-101: done

pkg/translator/listener.go
ziyue-101: I decided to have this pr to support same namespace reference for now. a cross namespace check will be added as a followup. I added a TODO in the code

pkg/translator/listener.go
ziyue-101: done

pkg/translator/listener.go
ziyue-101: added new struct constructors.

pkg/translator/translator.go
ziyue-101: done

pkg/translator/translator.go
ziyue-101: updated to log a warning for now. Whether to support >1 certs is TBD.  I added a TODO in the code and will address it in a followup.

pkg/translator/translator.go
ziyue-101: I decided to have this pr to support same namespace reference for now. a cross namespace check will be added as a followup. I added a TODO in the code

pkg/translator/translator.go
ziyue-101: done

pkg/translator/translator.go
ziyue-101: added `resolveCertificate` helper

pkg/translator/translator.go
ziyue-101: same, the code default to gateway namespace for now.

tests/e2e/controller_test.go
ziyue-101: done, add a new test pod without mounted certs

tests/e2e/controller_test.go
ziyue-101: done

tests/e2e/tls_helpers.go
ziyue-101: done

ziyue-101:
Discussed with @LiorLieberman  about the [problem](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/94#issuecomment-4300117469) 

I will revise the logic to "program" Gateway when the "FrontendValidation" field is missing to be able to stay conformant with Gateway API. 

I will also add documentation to ask users to provide both fields at the same time for mTLS.

pkg/translator/listener.go
```diff
@@ -163,64 +165,154 @@ func (t *Translator) validateCertificateRefs(gateway *gatewayv1.Gateway, listene
...
+}
+
+func (t *Translator) validateCACertificateRef(gateway *gatewayv1.Gateway, caRef gatewayv1.ObjectReference) *metav1.Condition {
+	ns := gateway.Namespace
+	if caRef.Namespace != nil {
+		ns = string(*caRef.Namespace)
+	}
+	configMapName := string(caRef.Name)
+
+	if caRef.Kind != "" && caRef.Kind != "ConfigMap" {
```
haiyanmeng: `caRef.Kind != ""` seems unnecessary.

pkg/translator/listener.go
```diff
@@ -163,64 +165,154 @@ func (t *Translator) validateCertificateRefs(gateway *gatewayv1.Gateway, listene
...
+		return &metav1.Condition{
+			Type:               string(gatewayv1.ListenerConditionResolvedRefs),
+			Status:             metav1.ConditionFalse,
+			Reason:             string(gatewayv1.ListenerReasonInvalidCACertificateKind),
+			Message:            fmt.Sprintf("CACertificateRef kind %s is not supported, only ConfigMap", caRef.Kind),
```
haiyanmeng: The message should be consistent with the message on Line 191.

pkg/translator/listener.go
```diff
@@ -163,64 +165,154 @@ func (t *Translator) validateCertificateRefs(gateway *gatewayv1.Gateway, listene
...
+			ObservedGeneration: gateway.Generation,
+		}
+	}
+
+	if ns != gateway.Namespace && !AllowedByReferenceGrant(gateway.Namespace, gatewayv1.GroupName, "Gateway", ns, "", "ConfigMap", configMapName, t.referenceGrantLister) {
+		return &metav1.Condition{
+			Type:               string(gatewayv1.ListenerConditionResolvedRefs),
+			Status:             metav1.ConditionFalse,
+			Reason:             string(gatewayv1.ListenerReasonRefNotPermitted),
+			Message:            fmt.Sprintf("Reference to ConfigMap %s in namespace %s not allowed by ReferenceGrant", configMapName, ns),
```
haiyanmeng: The message should be consistent with the message on Line 262.

LiorLieberman:
Worth mentioning the part we discussed about https://github.com/kubernetes-sigs/kube-agentic-networking/issues/254 as the longer term solution

pkg/translator/common_test.go
```diff
@@ -26,8 +26,9 @@ import (
 	agenticv0alpha0 "sigs.k8s.io/kube-agentic-networking/api/v0alpha0"
 )
 
-func newTestGateway(name, ns string) *gatewayv1.Gateway {
-	return &gatewayv1.Gateway{
+//nolint:unparam
```
LiorLieberman, line 31: whats the reason we need this?
relevant for all other `//nolint:unparam` comments

pkg/translator/translator.go
```diff
@@ -379,10 +370,215 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
...
+				}
+			}
+		}
+	}
+	return routesByListener, httpRouteStatuses
+}
+
+// extractCertificateRefs extract certificate references from gateway and listener
+// It returns the first certificate reference and ca reference, and whether there are multiple certificate references and ca references.
+// TODO: we do not support multiple certificate references and ca references for now.
```
haiyanmeng: This does not read like a TODO. Do we plan to support multiple certificate references and CA references in the future?

pkg/translator/translator.go
```diff
@@ -379,10 +370,215 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
 			resourcev3.ListenerType: finalEnvoyListeners,
 			resourcev3.RouteType:    envoyRoutes,
 			resourcev3.ClusterType:  clustersSlice,
+			resourcev3.SecretType:   envoySecrets,
 		}, orderedStatuses,
 		httpRouteStatuses, nil, nil
 }
 
+func (t *Translator) getAndGroupHTTPRoutes(gateway *gatewayv1.Gateway) (
```
LiorLieberman: func is returning routes and statuses, maybe select a different name? Also dont recommend using "get" name for methods in go. 

Something like `HTTPRoutesAndStatuses`?

site-src/guides/mtls.md
```diff
@@ -0,0 +1,69 @@
+# Mutual TLS (mTLS) Configuration
```
LiorLieberman, line 1: This guide have to be referenced from other guide/quickstart, as without mtls some guides wont work.

LiorLieberman (commented):
looks good to me overall, thanks @ziyue-101! 
Left a few comments.

pkg/translator/common_test.go
ziyue-101: Added a comment.

I did not hardcode the value into the test helper as i'd image some future use for the namespace variable

site-src/guides/mtls.md
ziyue-101: what other guide/quickstart are you thinking about?

For quickstart, it works as it because it is supposed to be the simplest set up with default mTLS, 

This guide is intended for users who wants to set up custom mTLS.

pkg/translator/translator.go
ziyue-101: done

pkg/translator/translator.go
ziyue-101: removed the TODO, for now just restrict to one reference. we  should access the use case of >1 certificate in the future.

ziyue-101:
> Worth mentioning the part we discussed about #254 as the longer term solution

Added to the doucumentation.

LiorLieberman (approved):
Thanks Ziyue! Looks good to me.

Approving and will let @haiyanmeng lgtm

PR #234, opened 2026-04-08
fix(controller): cap workqueue retries and RetryOnConflict on writes

### LINKED ISSUE #100: Holistic retry in controller
### Description
The current KAN controller implementation uses basic client-go workqueue plumbing but lacks a robust, holistic approach to error handling and retries. This leads to two main issues
1. Silent Failures: Certain errors (notably in updateGatewayStatus) are logged but not returned, preventing the workqueue from triggering a backoff and retry.
2. Infinite Loops: The controller does not track retry counts, meaning a permanently broken resource will be retried indefinitely, consuming CPU and generating noisy logs.

### Proposed Changes

Enhance the Worker Loop by adopting a holistic retry approach. This means
-  syncHandler and its sub-functions return all errors and stays agnostic of retry counts. 
-  The Worker Loop then becomes the 'brain' that decides whether to apply exponential backoff via `AddRateLimited`, or to permanently drop a resource after checking `NumRequeues`. 
- Wrap all Update and UpdateStatus calls in `retry.RetryOnConflict` to ensure the latest version of an object is always re-fetched from the lister inside retry/update blocks to avoid stale resourceVersion errors.

HarshithaMS005:
**What type of PR is this?**
/kind feature

- Gateway workqueue: On syncGateway error, requeue with AddRateLimited only while NumRequeues(key) < 15; otherwise Forget and log (stops unbounded retries).
- Backend finalizer workqueue: Same pattern with cap 15 for syncBackendFinalizer.
- Gateway writes: Finalizer add/remove and UpdateStatus go through retry.RetryOnConflict, re-reading the Gateway from the lister inside each attempt.
- GatewayClass writes: Finalizer add/remove and UpdateStatus use retry.RetryOnConflict + lister read; on successful status update, still calls enqueueGatewaysForClass.
- XBackend finalizer: Add/remove finalizer Update wrapped in retry.RetryOnConflict + lister read.
- XAccessPolicy status: UpdateStatus wrapped in retry.RetryOnConflict; each attempt loads the policy from the lister; ObservedGeneration uses that copy’s Generation.

Fixes #100 

```release-note
 NONE
```

### DIFF
--- a/pkg/controller/accesspolicy.go
+++ b/pkg/controller/accesspolicy.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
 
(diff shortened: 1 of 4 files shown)


haiyanmeng (approved):
/lgtm
/approve 

Thanks, @HarshithaMS005 !

PR #236, opened 2026-04-09
fix: update XAccessPolicy status when rules are skipped during translation

### LINKED ISSUE #233: Update XAccessPolicy status when rules are skipped during translation
## Description
Currently, when the translator encounters a semantically invalid rule in an `XAccessPolicy` (which passed API server CEL validations), it logs an error and skips the rule. However, it does not update the status of the `XAccessPolicy` to reflect this failure. 
This leaves the policy in a misleading `Accepted: True` state in status, while it is not fully applied in the data plane (Envoy).
### Examples of current silent failures:
1. **Unsupported backend kind for ExternalAuth**: In `pkg/translator/listener.go:509`, if an `ExternalAuth` rule uses the HTTP protocol but targets a backend that is not a `Service`, the translator logs an error and skips the rule.
2. **Unique ID generation failure**: In `pkg/translator/accesspolicy.go:272`, if `externalAuthUniqueID` fails, the rule is skipped.
## Expected Behavior
If a policy contains rules that cannot be translated, its status should be updated to reflect that it is invalid or partially applied. Following Gateway API conventions, it should ideally be marked with `PolicyConditionAccepted: False` and reason `PolicyReasonInvalid`.
## Proposed Solutions
We should implement one of the following approaches to improve observability:
1. **Pre-validation in Controller**: Move semantic checks (like validating that HTTP external auth targets a Service) into the controller's reconciliation loop. This would allow the controller to set `Accepted: False` with reason `Invalid` directly.
2. **Translation Feedback**: Allow the translator to return translation errors/conditions back to the controller, which can then update the policy status accordingly.

HarshithaMS005:
**What type of PR is this?**
/kind feature

This PR implements [[translation feedback]](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/233) so the controller no longer leaves XAccessPolicy in a misleading Accepted: True state when the translator skips or only partially applies rules that still pass API validation.
Today, failures such as HTTP ExternalAuth with a non-Service backend or externalAuthUniqueID errors are only logged; Envoy may omit behaviour while status still looks healthy. [Issue #233](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/233) asks for Gateway API–style policy conditions: Accepted: False with reason Invalid (PolicyReasonInvalid).

Fixes #233 

```release-note
NONE
```

### DIFF
--- a/pkg/controller/accesspolicy.go
+++ b/pkg/controller/accesspolicy.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -39,28 +40,7 @@ import (
 	agenticinformersv1alpha1 "sigs.k8s.io/kube-agentic-networking/k8s/client/informers/externalversions/api/v1alpha1"
 	"sigs.k8s.io/kube-agentic-networking/pkg/constants"
 	helpersv1alpha1 "sigs.k8s.io/kube-agentic-networking/pkg/helpers"
-	"sigs.k8s.io/kube-agentic-networking/pkg/translator"
 )
-
-// AccessPolicyTargetRefIndex is the index name for looking up AccessPolicies by target ref (namespace/name of XBackend).
-const AccessPolicyTargetRefIndex = "targetRef"
-
-// accessPolicyTargetRefIndexFunc indexes AccessPolicies by each XBackend targetRef (namespace/name).
-// Used by the informer cache to support O(1) lookup of policies targeting a given backend.
-func accessPolicyTargetRefIndexFunc(obj interface{}) ([]string, error) {
-	policy, ok := obj.(*agenticv1alpha1.XAccessPolicy)
-	if !ok {
-		return nil, nil
-	}
-	var keys []string
-	for _, targetRef := range policy.Spec.TargetRefs {
-		if !isXBackendTargetRef(targetRef) {
-			continue
-		}
-		keys = append(keys, policy.Namespace+"/"+string(targetRef.Name))
-	}
-	return keys, nil
-}
 
 func (c *Controller) setupAccessPolicyEventHandlers(accessPolicyInformer agenticinformersv1alpha1.XAccessPolicyInformer) error {
 	_, err := accessPolicyInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
@@ -80,11 +60,6 @@ func (c *Controller) onAccessPolicyAdd(obj interface{}) {
 		return
 	}
 
-	// Validate CEL expressions.
-	if !c.validateCELSpec(context.Background(), policy) {
-		return
-	}
-
 	c.enqueueGatewaysForAccessPolicy(policy)
 }
 
@@ -100,11 +75,6 @@ func (c *Controller) onAccessPolicyUpdate(old, newObj interface{}) {
 			if !c.isPolicyUnderTargetLimit(context.Background(), newPolicy) {
 				return
 			}
-		}
-
-		// Validate CEL expressions.
-		if !c.validateCELSpec(context.Background(), newPolicy) {
-			return
 		}
 
 		c.enqueueGatewaysForAccessPolicy(newPolicy)
@@ -266,7 +236,7 @@ func (c *Controller) isPolicyUnderTargetLimit(ctx context.Context, policy *agent
 	// 3. Update status for all targets based on the overall result.
 	for _, targetRef := range policy.Spec.TargetRefs {
 		reason := agenticv1alpha1.PolicyReasonAccepted
-		message := "AccessPolicy accepted for target"
+		message := accessPolicyAcceptedMessage
 
 		if !shouldAccept {
 			reason = agenticv1alpha1.PolicyLimitPerTargetExceeded
@@ -363,31 +333,78 @@ func (c *Controller) updateAccessPolicyStatus(ctx context.Context, policy *agent
 	})
 }
 
-// validateCELSpec validates that all CEL expressions in the policy are syntactically valid.
-// It returns true if all expressions are valid, false otherwise.
-// It also updates the policy status accordingly.
-func (c *Controller) validateCELSpec(ctx context.Context, policy *agenticv1alpha1.XAccessPolicy) bool {
-	if policy.Spec.Rules == nil {
-		return true
-	}
-
-	for _, rule := range policy.Spec.Rules {
-		if rule.Authorization != nil && rule.Authorization.Type == agenticv1alpha1.AuthorizationRuleTypeCEL && rule.Authorization.CEL != nil {
-			if _, err := translator.CompileCelExpression(rule.Authorization.CEL.Expression); err != nil {
-				msg := fmt.Sprintf("Failed to compile CEL expression %q: %v", rule.Authorization.CEL.Expression, err)
-				c.rejectPolicyForAllTargets(ctx, policy, msg)
-				return false
-			}
+// filterEmptyStrings returns a copy of ss without "" elements.
+func filterEmptyStrings(ss []string) []string {
+	var out []string
+	for _, s := range ss {
+		if s != "" {
+			out = append(out, s)
 		}
 	}
-
-	return true
+	return out
 }
 
-func (c *Controller) rejectPolicyForAllTargets(ctx context.Context, policy *agenticv1alpha1.XAccessPolicy, message string) {
-	for _, targetRef := range policy.Spec.TargetRefs {
-		if err := c.updateAccessPolicyStatus(ctx, policy, targetRef, false, agenticv1alpha1.PolicyReasonInvalidCEL, message); err != nil {
-			runtime.HandleError(fmt.Errorf("failed to update AccessPolicy status: %w", err))
+const (
+	accessPolicyAcceptedMessage    = "AccessPolicy accepted for target"
+	accessPolicySkippedRulesPrefix = "some rules were not programmed"
+	accessPolicyStaleInvalidReason = string(gwapiv1.PolicyReasonInvalid)
+)
+
+func accessPolicyNeedsTranslationStatusRefresh(policy *agenticv1alpha1.XAccessPolicy) bool {
+	for _, anc := range policy.Status.Ancestors {
+		cond := meta.FindStatusCondition(anc.Conditions, string(agenticv1alpha1.PolicyConditionAccepted))
+		if cond == nil {
+			continue
+		}
+		// Clear status left by older reconciliations that marked the whole policy Invalid.
+		if cond.Status == metav1.ConditionFalse && cond.Reason == accessPolicyStaleInvalidReason {
+			return true
+		}
+		if cond.Status == metav1.ConditionTrue && strings.Contains(cond.Message, accessPolicySkippedRulesPrefix) {
+			return true
+		}
+	}
+	return false
+}
+
+func accessPolicyTranslationWarningMessage(skippedRuleMsgs []string) string {
+	return fmt.Sprintf("%s; %s: %s", accessPolicyAcceptedMessage, accessPolicySkippedRulesPrefix,
+		strings.Join(deduplicateStrings(filterEmptyStrings(skippedRuleMsgs)), "; "))
+}
+
+// reconcileAccessPolicyTranslationStatus updates XAccessPolicy status when the translator skips rules.
+// Skipped rules are treated as "never match" at enforcement time; the policy stays Accepted so it
+// remains attached (default deny for allow policies when no rules match). Valid rules in the same
+// policy continue to be enforced.
+//
+// ListAccessPoliciesAttachedToGateway defines which policies are in scope for this Gateway sync.
+// The translator's issues map only contains policies that reported errors this pass; policies that
+// now translate cleanly but still carry a skipped-rules warning (or stale Invalid from older code)
+// are reset via isPolicyUnderTargetLimit when issues omits the policy.
+func (c *Controller) reconcileAccessPolicyTranslationStatus(ctx context.Context, gateway *gwapiv1.Gateway, issues map[types.NamespacedName][]string) {
+	attached, err := c.translator.ListAccessPoliciesAttachedToGateway(gateway)
+	if err != nil {
+		runtime.HandleError(fmt.Errorf("list AccessPolicies attached to gateway: %w", err))
+		return
+	}
+	for _, policy := range attached {
+		nn := types.NamespacedName{Namespace: policy.Namespace, Name: policy.Name}
+		if msgs, hasErr := issues[nn]; hasErr {
+			if !helpersv1alpha1.IsXAccessPolicyAccepted(policy) {
+				if !c.isPolicyUnderTargetLimit(ctx, policy) {
+					continue
+				}
+			}
+			msg := accessPolicyTranslationWarningMessage(msgs)
+			for _, tr := range policy.Spec.TargetRefs {
+				if err := c.updateAccessPolicyStatus(ctx, policy, tr, true, agenticv1alpha1.PolicyReasonAccepted, msg); err != nil {
+					runtime.HandleError(fmt.Errorf("update AccessPolicy %s status for skipped rules: %w", nn, err))
+				}
+			}
+			continue
+		}
+		if accessPolicyNeedsTranslationStatusRefresh(policy) {
+			_ = c.isPolicyUnderTargetLimit(ctx, policy)
 		}
 	}
 }
--- a/pkg/controller/accesspolicy_test.go
+++ b/pkg/controller/accesspolicy_test.go
@@ -18,17 +18,24 @@ package controller
 
 import (
 	"context"
+	"reflect"
 	"testing"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewayclientfake "sigs.k8s.io/gateway-api/pkg/client/clientset/versioned/fake"
+	gatewayinformers "sigs.k8s.io/gateway-api/pkg/client/informers/externalversions"
 
 	agenticv0alpha0 "sigs.k8s.io/kube-agentic-networking/api/v0alpha0"
 	agenticv1alpha1 "sigs.k8s.io/kube-agentic-networking/api/v1alpha1"
 	agenticclient "sigs.k8s.io/kube-agentic-networking/k8s/client/clientset/versioned/fake"
 	agenticinformers "sigs.k8s.io/kube-agentic-networking/k8s/client/informers/externalversions"
+	"sigs.k8s.io/kube-agentic-networking/pkg/constants"
+	"sigs.k8s.io/kube-agentic-networking/pkg/translator"
 )
 
 func TestIsPolicyUnderTargetLimit(t *testing.T) {
@@ -220,7 +227,7 @@ func TestIsPolicyUnderTargetLimit(t *testing.T) {
 			allPolicies := make([]runtime.Object, 0, len(tt.existing)+1)
 			allPolicies = append(allPolicies, tt.existing...)
 			allPolicies = append(allPolicies, tt.currentPolicy)
-			fakeClient := agenticclient.NewSimpleClientset(allPolicies...)
+			fakeClient := agenticclient.NewClientset(allPolicies...)
 			informerFactory := agenticinformers.NewSharedInformerFactory(fakeClient, 0)
 			lister := informerFactory.Agentic().V1alpha1().XAccessPolicies().Lister()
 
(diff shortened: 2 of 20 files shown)


haiyanmeng (commented):
@HarshithaMS005 , thanks for the PR!

Can you add unit test for this change?

pkg/translator/accesspolicy.go
```diff
@@ -269,6 +269,7 @@ func (t *Translator) translateAccessPolicyToRBAC(accessPolicy *agenticv0alpha0.X
 					hash, err := externalAuthUniqueID(rule.Authorization.ExternalAuth)
 					if err != nil {
 						klog.Errorf("Failed to generate unique ID for externalAuth config in AccessPolicy %s/%s: %v", accessPolicy.Namespace, accessPolicy.Name, err)
+						apTranslation.record(accessPolicy.Namespace, accessPolicy.Name, fmt.Sprintf("rule %q: cannot fingerprint external authorization config: %v", rule.Name, err))
```
guicassolato: Have you been able to test this error recorded to the status? I don't think I've ever witnessed it happening in practice.

pkg/controller/accesspolicy.go
```diff
@@ -278,56 +278,113 @@ func getTargetID(ref gwapiv1.LocalPolicyTargetReferenceWithSectionName) string {
...
+	for _, ap := range attached {
+		policy, err := c.agentic.accessPolicyLister.XAccessPolicies(ap.Namespace).Get(ap.Name)
+		if err != nil {
+			continue
+		}
+		nn := types.NamespacedName{Namespace: policy.Namespace, Name: policy.Name}
+		if msgs, hasErr := issues[nn]; hasErr {
+			msg := strings.Join(dedupeStringsPreserveOrder(msgs), "; ")
+			for _, tr := range policy.Spec.TargetRefs {
+				fresh, getErr := c.agentic.accessPolicyLister.XAccessPolicies(policy.Namespace).Get(policy.Name)
```
guicassolato: I believe `updateAccessPolicyStatus` already reads the policy fresh, doesn't it?

https://github.com/kubernetes-sigs/kube-agentic-networking/blob/bbeb9d74ceda30cc5f513372b12ca99049743489/pkg/controller/accesspolicy.go#L282

HarshithaMS005: @guicassolato updateAccessPolicyStatus does not call Get; it only DeepCopys the policy you pass in. The inner Get is so each targetRef runs against the latest status and resourceVersion after the previous UpdateStatus (needed when there are multiple targets).

pkg/translator/accesspolicy.go
HarshithaMS005: Good catch — we have not seen json.Marshal fail on a real HTTPExternalAuthFilter in practice. That branch is defensive. Now i added TestTranslateAccessPolicyToRBAC_recordsExternalAuthFingerprintFailure, which stubs the marshal hook to force the error and checks the exact collector message that becomes status. The user-visible case we do hit in real configs is still the HTTP non-Service backend path (TestBuildExtAuthzRecordsTranslationIssueForHTTPNonServiceBackend / full translate test).

pkg/controller/accesspolicy.go
guicassolato: Here we call `updateAccessPolicyStatus` for each targetRef. The first thing `updateAccessPolicyStatus` does (all wrapped within a retry block) is getting the policy fresh (line 283 in this branch). That line looks identical to this one (380).

I'm sorry. What am I missing?

HarshithaMS005: You’re right — I was answering the pre-retry version of updateAccessPolicyStatus. On this branch the first step inside RetryOnConflict is already accessPolicyLister.Get(policy.Namespace, policy.Name), so the Get in reconcileAccessPolicyTranslationStatus was duplicate. I’ve dropped the outer Get and kept a single call to updateAccessPolicyStatus per targetRef, thanks for catching it.

pkg/controller/accesspolicy.go
```diff
@@ -331,3 +332,59 @@ func (c *Controller) updateAccessPolicyStatus(ctx context.Context, policy *agent
 		return err
 	})
 }
+
+func dedupeStringsPreserveOrder(ss []string) []string {
+	seen := make(map[string]struct{})
+	var out []string
+	for _, s := range ss {
+		if s == "" {
```
bowei: This method seems to have hidden behavior that is unexpected (skips empty strings). This is not documented in a docstring and is unexpected. "" is still a valid string. 

If we are implementing a generic method, we should make its contract very clear and simple.

If you want to filter out empty strings, I suggest you do it outside of this func:

something like `dedupeStrings(filterEmptyString(ss))`

Suggest:
 
```
func dedupeStringsPreserveOrder(ss []string) []string {
  seen := make(map[string]struct{})
  var out []string
  for _, s := range ss {
    if _, ok := seen[s]; !ok {
      out = append(out, s)
      seen[s] = struct{}{}
    }
  }
  return out
}
```

pkg/translator/listener.go
```diff
@@ -686,8 +706,10 @@ func toEnvoyExactStringMatchers(s []string) []*matcherv3.StringMatcher {
 	return matchers
 }
 
+var marshalExternalAuthForUniqueID = json.Marshal
```
bowei, line 865: you created a global var here but don't document why it is global. 

If it is for error injection or testing, you need to explicitly comment it as such. Also, this may run in races when the tests are run in parallel? Have you checked that?

pkg/translator/translator.go
```diff
@@ -63,6 +63,38 @@ func (e *ControllerError) Error() string {
 	return e.Message
 }
 
+// accessPolicyTranslationCollector records semantic translation failures for
+// XAccessPolicy resources during a single Gateway translation. The controller
+// uses the snapshot to set Policy Accepted=False with PolicyReasonInvalid.
+type accessPolicyTranslationCollector struct {
```
bowei: Do we want to do a collector for just access policies or make this more generic and have it collect errors in general, but for the first implementation, collect access policy errors


```
type translationErrors struct {
  accessPolicyIssues map[...][]error
}
```

This way other translation errors can also use this struct accordingly

pkg/translator/translator.go
```diff
@@ -63,6 +63,38 @@ func (e *ControllerError) Error() string {
...
+	issues map[types.NamespacedName][]string
```
bowei: Does it make more sense for the collected errors to not be string but Go `error` type?

pkg/translator/listener.go
```diff
@@ -443,6 +461,7 @@ func (t *Translator) buildExtAuthzFilters(gateway *gatewayv1.Gateway) ([]*hcm.Ht
 			uniqueID, err := externalAuthUniqueID(extAuthz)
 			if err != nil {
 				klog.Error(err)
+				apTranslation.record(apKey.Namespace, apKey.Name, fmt.Sprintf("rule %q: cannot fingerprint external authorization config: %v", rule.Name, err))
```
bowei: I would prefer instead of sprinkling fmt.Sprintf(...) all throughout the code base that we have a more structured error collector interface:

```
func (c *...Collector) record(nsname types.NamespacedName, desc string, err error)

apTranslation.record(ap.Namespace, ap.Name, "cannot fingerprint external auth config", err)
```

pkg/translator/listener.go
```diff
@@ -435,6 +452,7 @@ func (t *Translator) buildExtAuthzFilters(gateway *gatewayv1.Gateway) ([]*hcm.Ht
 		if !t.isAccessPolicyAttachedToGateway(ap, gateway) {
 			continue
 		}
+		apKey := types.NamespacedName{Namespace: ap.Namespace, Name: ap.Name}
```
bowei: I can't find any reference to apKey being used other than to use its fields directly. Why did you create this variable?

pkg/translator/listener.go
bowei: This comment applies to any globals you have created that ordinarily should not be globals.

pkg/controller/accesspolicy.go
haiyanmeng: We already have https://github.com/kubernetes-sigs/kube-agentic-networking/blob/ca5be6fb548e08069b7f19560def222c1d12333c/pkg/controller/referencegrant.go#L209

pkg/controller/accesspolicy_test.go
```diff
@@ -224,3 +231,237 @@ func TestIsPolicyUnderTargetLimit(t *testing.T) {
...
+	}
+}
+
+func TestReconcileAccessPolicyTranslationStatus_setsInvalid(t *testing.T) {
+	ctx := context.Background()
+	ns := "default"
+	gwName := "gw1"
+	policy := testPolicyGatewayTarget("pol1", ns, gwName)
+
+	fakeAgentic := agenticclient.NewSimpleClientset(policy)
```
haiyanmeng: NewSimpleClientset is deprecated. Use NewClientset

pkg/translator/accesspolicy_test.go
```diff
@@ -749,6 +752,55 @@ func TestTranslateAccessPolicyToRBAC(t *testing.T) {
...
+	snap := coll.snapshot()
+	nn := types.NamespacedName{Namespace: "default", Name: "pol-fingerprint"}
+	msgs, ok := snap[nn]
+	if !ok || len(msgs) != 1 {
+		t.Fatalf("expected one recorded issue for %v, got %#v", nn, snap)
+	}
+	if !strings.Contains(msgs[0], `rule "ext-rule": cannot fingerprint`) {
+		t.Errorf("message %q should mention rule and fingerprint", msgs[0])
+	}
+	if !strings.Contains(msgs[0], "simulated marshal failure") {
```
haiyanmeng: Define this as an constant

pkg/controller/accesspolicy.go
```diff
@@ -331,3 +332,59 @@ func (c *Controller) updateAccessPolicyStatus(ctx context.Context, policy *agent
...
+		}
+	}
+	return false
+}
+
+// reconcileAccessPolicyTranslationStatus updates XAccessPolicy status when the translator skips or
+// partially applies rules (Gateway API PolicyReasonInvalid), and clears that reason by re-running
+// the per-target limit check when translation succeeds again.
+func (c *Controller) reconcileAccessPolicyTranslationStatus(ctx context.Context, gateway *gwapiv1.Gateway, issues map[types.NamespacedName][]string) {
+	attached, err := c.translator.ListAccessPoliciesAttachedToGateway(gateway)
```
haiyanmeng, line 385: Let us add a comment to explain why `c.translator.ListAccessPoliciesAttachedToGateway(gateway)` is needed

pkg/translator/listener.go
```diff
@@ -158,13 +159,29 @@ func (t *Translator) validateListeners(gateway *gatewayv1.Gateway) map[gatewayv1
 	return listenerConditions
 }
 
-func (t *Translator) translateListenerToFilterChain(lis gatewayv1.Listener, routeName string, gateway *gatewayv1.Gateway) (*listener.FilterChain, error) {
+// ListAccessPoliciesAttachedToGateway returns every XAccessPolicy (any namespace) that is
+// attached to the Gateway via a Gateway or XBackend targetRef reachable from this Gateway's HTTPRoutes.
+func (t *Translator) ListAccessPoliciesAttachedToGateway(gateway *gatewayv1.Gateway) ([]*v0alpha0.XAccessPolicy, error) {
+	all, err := t.accessPolicyLister.List(labels.Everything())
```
haiyanmeng, line 342: This list seems expensive. Can we improve it?

pkg/translator/translator.go
haiyanmeng: Add comments to explain the key and value of the map

pkg/translator/listener.go
HarshithaMS005: global marshalExternalAuthForUniqueID — This exists only so unit tests can inject a marshal failure for externalAuthUniqueID; production always behaves like json.Marshal. Documented on the var that production must not reassign it. Parallel safety: only one test overrides it, it restores via t.Cleanup, and the package’s tests don’t use t.Parallel(); the doc warns future tests not to combine parallel execution with this hook.

pkg/controller/accesspolicy.go
HarshithaMS005: Thank you for pointing it.

pkg/translator/translator.go
HarshithaMS005: Thanks — agreed. I refactored the one-off collector into a translationErrors struct with an accessPolicyIssues field for this PR. The type comment calls out that we can add more maps (e.g. other resources) in later changes without growing the TranslateGatewayToXDS parameter list.

pkg/translator/translator.go
HarshithaMS005: Good call. The accumulator now stores []error per XAccessPolicy (using fmt.Errorf with %w where we have an underlying error so unwrap works).

pkg/translator/listener.go
HarshithaMS005: Sure

pkg/translator/listener.go
HarshithaMS005: We introduced apKey because recordAccessPolicyIssue takes types.NamespacedName. It’s passed whole on both fingerprint and filter-build failures (not only .Namespace/.Name). Before that refactor the helper took two strings, so the temp looked useless.

pkg/controller/accesspolicy_test.go
HarshithaMS005: okay sure

pkg/translator/accesspolicy_test.go
HarshithaMS005: Sure

pkg/controller/accesspolicy.go
HarshithaMS005: Okay sure.

pkg/translator/listener.go
HarshithaMS005: Thanks — agreed that listing every XAccessPolicy in the cluster on each Gateway sync didn’t scale.

We improved it in two ways:

- Informer indexes – The AccessPolicy informer now registers a gatewayTarget index (policies whose target is a Gateway: key policy.Namespace/gatewayName) alongside the existing targetRef index for XBackend targets (policy.Namespace/backendName). For a given Gateway we union policies from the Gateway index with policies looked up by each XBackend backendRef on HTTPRoutes attached to that Gateway (same attachment rules as isAccessPolicyAttachedToGateway), deduped by UID.
- Reuse – buildExtAuthzFilters now calls ListAccessPoliciesAttachedToGateway instead of listing all policies and filtering again.

The controller passes accessPolicyInformer.GetIndexer() into the translator so production uses the indexed path; tests that pass a nil indexer keep the previous list-and-filter behaviour.

pkg/translator/accesspolicy_index.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	agenticv0alpha0 "sigs.k8s.io/kube-agentic-networking/api/v0alpha0"
+)
+
+const (
+	// AccessPolicyTargetRefIndex indexes XAccessPolicies by XBackend targetRef key "namespace/name"
+	// (policy namespace + backend name). Must stay aligned with the controller informer registration.
+	AccessPolicyTargetRefIndex = "targetRef"
```
haiyanmeng: Let us change this to `AccessPolicyBackendTargetIndex = "backendTarget"` to be consistent with line 31

pkg/translator/listener.go
```diff
@@ -321,13 +322,85 @@ func (t *Translator) validateCACertificateRef(gateway *gatewayv1.Gateway, caRef
...
+	var out []*v0alpha0.XAccessPolicy
+	for _, ap := range all {
+		if t.isAccessPolicyAttachedToGateway(ap, gateway) {
+			out = append(out, ap)
+		}
+	}
+	return out, nil
+}
+
+func (t *Translator) listAccessPoliciesAttachedToGatewayIndexed(gateway *gatewayv1.Gateway) ([]*v0alpha0.XAccessPolicy, error) {
```
haiyanmeng: Why not use `isAccessPolicyAttachedToGateway` in this method?

pkg/translator/listener.go
```diff
@@ -321,13 +322,85 @@ func (t *Translator) validateCACertificateRef(gateway *gatewayv1.Gateway, caRef
 	return nil
 }
 
-func (t *Translator) translateListenerToFilterChain(lis gatewayv1.Listener, routeName string, gateway *gatewayv1.Gateway) (*listener.FilterChain, error) {
+// ListAccessPoliciesAttachedToGateway returns every XAccessPolicy that is attached to the Gateway
+// via a Gateway targetRef or an XBackend targetRef reachable from this Gateway's HTTPRoutes.
+//
+// When accessPolicyIndexer is set (controller wiring), this uses informer indexes instead of
+// listing every XAccessPolicy in the cluster.
+func (t *Translator) ListAccessPoliciesAttachedToGateway(gateway *gatewayv1.Gateway) ([]*v0alpha0.XAccessPolicy, error) {
```
haiyanmeng: Can we struct this method to sth similar to https://github.com/kubernetes-sigs/kube-agentic-networking/blob/9d6bbc10e13f7a30040ab48e45344ff387d6f6d8/pkg/controller/referencegrant.go#L81-L133?

pkg/translator/accesspolicy_index.go
```diff
@@ -0,0 +1,69 @@
...
+	// AccessPolicyGatewayTargetIndex indexes XAccessPolicies by Gateway targetRef key "namespace/name"
+	// (policy namespace + gateway name) for LocalPolicyTargetReference Gateway targets.
+	AccessPolicyGatewayTargetIndex = "gatewayTarget"
+)
+
+// AccessPolicyXBackendTargetRefIndexFunc is the cache indexer for AccessPolicyTargetRefIndex.
+func AccessPolicyXBackendTargetRefIndexFunc(obj interface{}) ([]string, error) {
```
bowei: naming: ...Func suffix usually means you are returning a go function, but in this case it's a list of string. Consider removing the Func suffix.

Does this func need to be exported? I would consider having this package export a

func NewAccessPolicyIndexer() cache.Indexer rather than exposing these functions directly.

pkg/translator/accesspolicy_index.go
```diff
@@ -0,0 +1,69 @@
...
+			continue
+		}
+		keys = append(keys, policy.Namespace+"/"+string(targetRef.Name))
+	}
+	return keys, nil
+}
+
+// AccessPolicyGatewayTargetRefIndexFunc is the cache indexer for AccessPolicyGatewayTargetIndex.
+// Keys match isAccessPolicyAttachedToGateway for direct Gateway targets (local ref → policy namespace).
+func AccessPolicyGatewayTargetRefIndexFunc(obj interface{}) ([]string, error) {
```
bowei: naming: ...Func suffix usually means you are returning a go function, but in this case it's a list of string. Consider removing the Func suffix.

Does this func need to be exported?

pkg/translator/translator.go
```diff
@@ -65,6 +66,52 @@ func (e *ControllerError) Error() string {
...
+	var combined error
+	if ruleName != "" {
+		combined = fmt.Errorf("rule %q: %s: %w", ruleName, desc, err)
+	} else {
+		combined = fmt.Errorf("%s: %w", desc, err)
+	}
+	e.accessPolicyIssues[nn] = append(e.accessPolicyIssues[nn], combined)
+}
+
+func (e *translationErrors) accessPolicyIssuesSnapshot() map[types.NamespacedName][]string {
```
bowei: [naming] accessPolicyIssuesSnapshot => policyIssues()

Remove redundant names

pkg/translator/translator.go
```diff
@@ -65,6 +66,52 @@ func (e *ControllerError) Error() string {
...
+	accessPolicyIssues map[types.NamespacedName][]error
+}
+
+func newTranslationErrors() *translationErrors {
+	return &translationErrors{accessPolicyIssues: make(map[types.NamespacedName][]error)}
+}
+
+// recordAccessPolicyIssue attaches a short description and wraps err for one XAccessPolicy issue.
+// ruleName is optional; when set, the formatted error begins with rule "<ruleName>": ...
+func (e *translationErrors) recordAccessPolicyIssue(nn types.NamespacedName, ruleName, desc string, err error) {
```
bowei: [naming] normalize terms, prefer more succinct terms: recordAccessPolicyIssue -> accessPolicyError

pkg/translator/accesspolicy_index.go
HarshithaMS005: Sure

pkg/translator/listener.go
HarshithaMS005: listAccessPoliciesAttachedToGatewaySlow - Lists every XAccessPolicy in the cluster, then keeps only those for which isAccessPolicyAttachedToGateway(ap, gateway) is true. The helper is the filter because there is no other way to narrow the set cheaply.

listAccessPoliciesAttachedToGatewayIndexed - Implements the same attachment definition without a full list:

- Gateway targetRef: ByIndex(AccessPolicyGatewayTargetIndex, namespace/gatewayName) returns policies that target that Gateway (equivalent to the Gateway branch in isAccessPolicyAttachedToGateway).

- Backend targetRef: Uses getHTTPRoutesForGateway (same as the helper), enumerates XBackend refs on those routes, and for each key namespace/backendName calls ByIndex(AccessPolicyBackendTargetIndex, key) (equivalent to the helper’s “policy targets this backend and that backend appears on a route for this Gateway” logic). So the indexed code is not omitting the helper’s semantics; it is a targeted lookup version of the same rules for performance. 
Under correct index functions, every policy returned by those lookups should already satisfy the helper. Re-filtering would duplicate work; the real invariant is index key logic stays aligned with isAccessPolicyAttachedToGateway

pkg/translator/listener.go
HarshithaMS005: Sure

pkg/translator/accesspolicy_index.go
HarshithaMS005: Thanks for the feedback — addressed in the latest push.
- Dropped the misleading Func suffix on the index implementations; they’re unexported as xAccessPolicyByBackendTarget and xAccessPolicyByGatewayTarget, and the comments note they implement cache.IndexFunc.
- Stopped exporting those helpers. The package now exports NewAccessPolicyIndexers() cache.Indexers (using the client-go Indexers type for AddIndexers), and the index name constants stay exported for ByIndex call sites. The controller uses AddIndexers(translator.NewAccessPolicyIndexers()).

pkg/translator/accesspolicy_index.go
HarshithaMS005: Agreed. Dropped the Func names; the index impls are unexported (xAccessPolicyBy*) and the only registration entry point is NewAccessPolicyIndexers(). The index string constants stay exported for ByIndex lookups.

pkg/translator/translator.go
HarshithaMS005: Addressed.

pkg/translator/translator.go
HarshithaMS005: Addressed

david-martin:
@HarshithaMS005 

Two questions as I read through this:

1. Should we reject the entire policy rather than partially enforcing it when a rule can't be translated? Partial enforcement of security policy means the effective behavior differs from what the user wrote, which could be a security gap. Gateway API takes an all-or-nothing approach with HTTPRoute. Curious what you and @guicassolato think.
2. For the ExternalAuth HTTP backend kind restriction, could we catch that earlier with a CEL validation rule on the CRD? The user would get immediate feedback from kubectl apply rather than waiting for the controller to reconcile.

Neither of these block the current PR, just want to make sure we've thought through the design before it sets a pattern.

LiorLieberman:
/hold
Will review today, thanks!

pkg/controller/accesspolicy.go
```diff
@@ -331,3 +312,60 @@ func (c *Controller) updateAccessPolicyStatus(ctx context.Context, policy *agent
...
+// reconciliation. Listing attachment lets us update both: set Invalid when issues[nn] is non-empty,
+// and clear stale Invalid (via isPolicyUnderTargetLimit) when issues omits the policy.
+func (c *Controller) reconcileAccessPolicyTranslationStatus(ctx context.Context, gateway *gwapiv1.Gateway, issues map[types.NamespacedName][]string) {
+	attached, err := c.translator.ListAccessPoliciesAttachedToGateway(gateway)
+	if err != nil {
+		runtime.HandleError(fmt.Errorf("list AccessPolicies attached to gateway: %w", err))
+		return
+	}
+	for _, ap := range attached {
+		policy, err := c.agentic.accessPolicyLister.XAccessPolicies(ap.Namespace).Get(ap.Name)
```
guicassolato: What's the goal of getting the policy object again? Is it to ensure the latest `spec.targetRefs` after getting state of the object from listing at line 347.

This seems almost the same as https://github.com/kubernetes-sigs/kube-agentic-networking/pull/236#discussion_r3093506511. I wonder if we this step is really needed.

If we can confirm this builds on the informer's cache and won't issue a request to the API server for each policy object unless the generation of the resource has changed, then perhaps there's no harm in keeping it. Otherwise, I'd probably advocate for avoiding the overhead on the API server, given another reconciliation event may have been enqueued already by the time this happens.

HarshithaMS005:
> Two questions as I read through this:
> 
> 1. Should we reject the entire policy rather than partially enforcing it when a rule can't be translated? Partial enforcement of security policy means the effective behavior differs from what the user wrote, which could be a security gap. Gateway API takes an all-or-nothing approach with HTTPRoute. Curious what you and @guicassolato think.
> 2. For the ExternalAuth HTTP backend kind restriction, could we catch that earlier with a CEL validation rule on the CRD? The user would get immediate feedback from kubectl apply rather than waiting for the controller to reconcile.
> 
> Neither of these block the current PR, just want to make sure we've thought through the design before it sets a pattern.

@david-martin Thanks for raising these — agreed they’re worth settling before this becomes the long-term pattern.
This PR only surfaces translation failures in XAccessPolicy status (Accepted=False, PolicyReasonInvalid) per #233. The translator still skips individual rules that fail translation and applies the rest; we haven’t changed that enforcement model here. I agree partial enforcement can diverge from what the user wrote and is worth a deliberate decision (Gateway API’s HTTPRoute precedent is a good reference). Happy to open a follow-up issue to discuss policy-level rejection vs rule-level skip with @guicassolato and the maintainers.

CEL for HTTP ExternalAuth backend kind: Agree that belongs in CRD validation where possible (we already use XValidation CEL on XAccessPolicy). I’d treat that as a separate API/CRD change so kubectl apply fails fast; this PR focuses on controller-reported Invalid for cases we can’t express in OpenAPI/CEL yet.

pkg/controller/accesspolicy.go
HarshithaMS005: @guicassolato Good catch — the re-Get isn’t needed for fresher spec.targetRefs. ListAccessPoliciesAttachedToGateway already returns full policies from the shared accessPolicyInformer lister (same store as c.agentic.accessPolicyLister). The lister Get is an in-memory cache lookup only (no API server read per policy). Status writes already re-fetch inside updateAccessPolicyStatus’s conflict retry before UpdateStatus.

I’ll drop the redundant Get and iterate attached directly unless you’d prefer keeping it for symmetry with other paths — functionally equivalent either way given the shared informer.

guicassolato:
> 1. Should we reject the entire policy rather than partially enforcing it when a rule can't be translated? Partial enforcement of security policy means the effective behavior differs from what the user wrote, which could be a security gap. Gateway API takes an all-or-nothing approach with HTTPRoute. Curious what you and @guicassolato think.

I would personally be in favour of rejecting the entire policy object.

It's a tricky one because a policy with many rules, that was working perfectly fine, could get completely invalidated because the user made a bad change to a single rule. Because of that, people could lose access or endpoints previously protected go completely open until the issue is fixed. However, I suppose that's what staging envs are for?

Arguably, for some people, managing policy status at the rule level is even worse of a situation to be.

I'll let others chime in.

pkg/translator/listener.go
```diff
@@ -309,13 +310,101 @@ func (t *Translator) validateCACertificateRef(gateway *gatewayv1.Gateway, caRef
...
+	}
+
+	if fallbackToListingAllAccessPolicies {
+		// Fallback to listing all AccessPolicies.
+		return t.listAccessPoliciesAttachedToGatewaySlow(gateway)
+	}
+	return policies, nil
+}
+
+func (t *Translator) listAccessPoliciesAttachedToGatewaySlow(gateway *gatewayv1.Gateway) ([]*v1alpha1.XAccessPolicy, error) {
```
guicassolato, line 341: listAccessPoliciesAttachedToGateway**Slow**. At least we're honest 😂

HarshithaMS005:
> I would personally be in favour of rejecting the entire policy object.
> 
> It's a tricky one because a policy with many rules, that was working perfectly fine, could get completely invalidated because the user made a bad change to a single rule. Because of that, people could lose access or endpoints previously protected go completely open until the issue is fixed. However, I suppose that's what stage ends are for?
> 
> Arguably, for some people, managing policy status at the rule level is even worse of a situation to be.
> 
> I'll let others chime in.

Thanks @david-martin and @guicassolato — really helpful framing.

Where this PR lands: it closes the gap from #233 where translation failures were only logged while status could still look healthy. When the translator records issues for a policy, we now set Accepted=False with PolicyReasonInvalid on that policy object (not per-rule conditions). That matches Gateway API’s policy-level status model and makes misconfiguration visible in kubectl.

What it does not do yet: we still skip the failing rule/config during xDS generation and may apply the rest of the same policy. So effective Envoy behavior can still differ from what the user wrote until they fix the spec — I agree that’s a security/consistency concern and aligns with the HTTPRoute all-or-nothing precedent @david-martin called out.

On rejecting the entire policy: I’m inclined the same way as @guicassolato — if any part of a policy fails translation, we should treat the whole policy as non-enforceable for that target (no partial RBAC/ext_authz from that object), while keeping status at the policy level. The “one bad edit invalidates a previously working policy” blast radius is real, but that’s what status + staging are for; rule-level status would be harder to reason about operationally.

I’d like to merge this PR as the status/visibility fix and track all-or-nothing enforcement as a follow-up (happy to open an issue and own it unless someone prefers we fold it in here). @guicassolato — does that split work for you?

pkg/translator/listener.go
HarshithaMS005: Yeah😄 — it’s the fallback when indexes aren’t available: list everything in the store and filter with isAccessPolicyAttachedToGateway. Happy to rename or add a one-line comment if that helps; functionally it’s only for attachment discovery for status updates, not for building partial config.

LiorLieberman:
Thanks for the discussion on this! To summarize and clarify our consensus (in meetings+slack) on how invalid/valid/partiallyvalid policy statuses and enforcement should behave :

* Invalid Rules: If a rule within a policy is invalid, it is treated as a "never match".

* Partial Validity: If a policy contains a mix of valid and invalid rules, the valid rules are still enforced.

* State Changes: If a previously valid rule becomes invalid, it transitions to a "never match". We will not preserve previous states, as this introduces too much complexity and won't survive controller restarts, downtime, or sharding.

Example: Suppose there is exactly one policy with a single valid rule (which implicitly means everything not allowed by this policy is denied by default). If that rule becomes invalid, it is treated as a policy with a rule that never matches. It is not treated as the absence of any allow policies.

On a separate note, I noticed the timeout is being bumped in this PR - could you clarify the reasoning behind that change?

HarshithaMS005:
> Thanks for the discussion on this! To summarize and clarify our consensus (in meetings+slack) on how invalid/valid/partiallyvalid policy statuses and enforcement should behave :
> 
> * Invalid Rules: If a rule within a policy is invalid, it is treated as a "never match".
> * Partial Validity: If a policy contains a mix of valid and invalid rules, the valid rules are still enforced.
> * State Changes: If a previously valid rule becomes invalid, it transitions to a "never match". We will not preserve previous states, as this introduces too much complexity and won't survive controller restarts, downtime, or sharding.
> 
> Example: Suppose there is exactly one policy with a single valid rule (which implicitly means everything not allowed by this policy is denied by default). If that rule becomes invalid, it is treated as a policy with a rule that never matches. It is not treated as the absence of any allow policies.
> 
> On a separate note, I noticed the timeout is being bumped in this PR - could you clarify the reasoning behind that change?

Thanks for summarizing the consensus — I've updated the PR to match.

**Enforcement**
- Invalid rules are skipped during translation (never match).
- Valid rules in the same policy are still enforced.
- Translation is stateless each gateway sync (no preserved “last good” RBAC).
- If the only rule becomes invalid, the policy stays attached with Accepted=True; it behaves like a policy whose rules never match (default deny for allow policies), not like the policy was removed.

**Status**
- Previously this PR set Accepted=False / Invalid when any rule was skipped, which caused findAccessPoliciesForTarget to drop the policy entirely. That’s fixed: skipped rules are surfaced as a warning on Accepted=True (some rules were not programmed). Stale Invalid status from earlier revisions is cleared on the next clean translation.

**CEL validation**
- Removed whole-policy validateCELSpec rejection that blocked reconciliation when one CEL rule was bad. Per-rule compile failures are handled in the translator (same as other translation errors).
- Gateway-level mergeAllowPoliciesToRBAC now reports skipped rules into the same translation-issues path.

**Timeout bump**
- The golangci timeout change (5m → 15m) was to fix flaky verify-golint on cold CI where go/packages load alone can approach 5m before linters run. It’s unrelated to access-policy behavior — I can move it to a separate commit/PR if you prefer.

PR #240, opened 2026-04-17
implement logic OR for access policies with the same target

### LINKED ISSUE #237: Change the evaluation rules for multiple InlineTools policies/rules targeting the same Gateway object
**Problem:**
Today, if there are multiple InlineTools policies/rules target the same Gateway object, the Envoy Agentic Gateway managed by our reference implementation allows a request only if all the policies/rules allow it. For example, we have two XAccessPolicy objects, `policy1` (which only allows an agent running with the Service Account `adk-agent-sa` to access `get-sum`) and `policy2` (which only allows an agent running with the Service Account `adk-agent-sa` to access `get-tiny-image`).  Currently, an agent running with the Service Account `adk-agent-sa` is not allowed to access both tools. 

```
apiVersion: agentic.prototype.x-k8s.io/v0alpha0
kind: XAccessPolicy
metadata:
  name: policy1
spec:
  targetRefs:
    ...  
  rules:
    - name: tools-for-adk-agent-sa
      source:
        type: ServiceAccount
        serviceAccount:
          name: adk-agent-sa
          namespace: quickstart-ns
      authorization:
        type: InlineTools
        tools:
          - "get-sum"
```

```
apiVersion: agentic.prototype.x-k8s.io/v0alpha0
kind: XAccessPolicy
metadata:
  name: policy2
spec:
  targetRefs:
    ...
  rules:
    - name: tools-for-adk-agent-sa
      source:
        type: ServiceAccount
        serviceAccount:
          name: adk-agent-sa
          namespace: quickstart-ns
      authorization:
        type: InlineTools
        tools:
          - "get-tiny-image"
```

**Desired behavior:**
if there are multiple InlineTools policies/rules target the same Gateway object, the Envoy Agentic Gateway managed by our reference implementation allows a request if any policies/rules allow it.

For the example above, an agent running with the Service Account `adk-agent-sa` should be allowed to access both tools.

ziyue-101:
**What type of PR is this?**

/kind feature

## Description
This PR implements logic **OR** semantics for `XAccessPolicy` rules targeting the same resource (Gateway or Backend). Previously, overlapping policies could result in sequential filters that blocked authorized traffic. 
Now, `InlineTools` rules from multiple policies are merged into a single Envoy RBAC filter. If `ExternalAuth` rules are also present, the merged `InlineTools` rules are copied into each `ExternalAuth` filter, ensuring that traffic is allowed if it satisfies *any* of the authorization conditions.
### Core Logic (`pkg/translator/`)
*   **Merged RBAC Filters**: Added `buildRBACConfigsForPolicies` in `accesspolicy.go` to handle the merging of `InlineTools` rules and their injection into `ExternalAuth` filters.
*   **Corrected Filter Counting**: Refactored `calculateMaxBackendRBACFilters` to compute the actual number of required filters after merging, preventing overestimation.
*   **Refactored Translation**: Extracted `translateAccessRuleToRBACPolicy` to handle mapping individual access rules to Envoy policies, improving code maintainability.
### Unit Tests (`pkg/translator/`)
*   **Reduced Filter Expectations**: Updated existing tests in `accesspolicy_test.go` and `listener_test.go` to account for the reduced filter count (policies are now grouped rather than generating one filter each).
*   **New Test Coverage**: Added new table tests covering mixed scenarios (inline + external auth) to ensure rules are properly merged across policies.
*   **Greened Verification Helpers**: Introduced `verifyRBACConfig` and `verifyRBACPerRouteConfig` to deduplicate assertion logic in tests.
### E2E Tests (`tests/e2e/`)
*   **Added Multi-Policy Cases**: Added "Case 6" (Gateways) and "Case 7" (Backends) to verify that creating multiple policies correctly results in accessible tools without blocking.
*   **New Test Artefacts**: Added 4 new YAML test data files to drive these specific test cases.

Fixes #237 

```release-note
Accesspolicies are OR'ed if they have the same target.
```

### DIFF
--- a/pkg/translator/accesspolicy.go
+++ b/pkg/translator/accesspolicy.go
@@ -68,15 +68,16 @@ func (t *Translator) buildGatewayLevelRBACFilters(gateway *gatewayv1.Gateway) ([
 		return nil, fmt.Errorf("failed to find Gateway access policies: %w", err)
 	}
 	var filters []*hcm.HttpFilter
-	for i, policy := range gwPolicies {
-		// Build the RBAC config for this policy
-		rbacProto := t.buildRBACConfigWithCommonPolicies(policy)
-		rbacAny, err := anypb.New(rbacProto)
+
+	rbacConfigs := t.buildRBACConfigsForPolicies(gwPolicies)
+
+	for _, rbacConfig := range rbacConfigs {
+		rbacAny, err := anypb.New(rbacConfig)
 		if err != nil {
 			return nil, err
 		}
 		filters = append(filters, &hcm.HttpFilter{
-			Name: fmt.Sprintf("%s%d", constants.GatewayRBACFilterNamePrefix, i+1),
+			Name: fmt.Sprintf("%s%d", constants.GatewayRBACFilterNamePrefix, len(filters)+1),
 			ConfigType: &hcm.HttpFilter_TypedConfig{
 				TypedConfig: rbacAny,
 			},
@@ -84,6 +85,82 @@ func (t *Translator) buildGatewayLevelRBACFilters(gateway *gatewayv1.Gateway) ([
 	}
 
 	return filters, nil
+}
+
+// buildRBACConfigsForPolicies creates RBAC configs for merged InlineTools and individual ExternalAuth rules.
+func (t *Translator) buildRBACConfigsForPolicies(policies []*agenticv0alpha0.XAccessPolicy) []*rbacv3.RBAC {
+	mergedInlineToolsRBAC := &rbacv3.RBAC{
+		Rules: &rbacconfigv3.RBAC{
+			Action:   rbacconfigv3.RBAC_ALLOW,
+			Policies: map[string]*rbacconfigv3.Policy{},
+		},
+	}
+	hasInlineToolsRules := false
+
+	var extAuthConfigs []*rbacv3.RBAC
+
+	for _, policy := range policies {
+		var extAuthRBAC *rbacv3.RBAC
+		hasExtAuthInThisPolicy := false
+
+		for _, rule := range policy.Spec.Rules {
+			rbacPolicy, isExternalAuth, extAuthStatPrefix, err := t.translateAccessRuleToRBACPolicy(policy.Namespace, rule)
+			if err != nil {
+				klog.Errorf("Failed to translate access rule %s in policy %s/%s: %v", rule.Name, policy.Namespace, policy.Name, err)
+				continue
+			}
+
+			if isExternalAuth {
+				hasExtAuthInThisPolicy = true
+				if extAuthRBAC == nil {
+					extAuthRBAC = &rbacv3.RBAC{
+						ShadowRules: &rbacconfigv3.RBAC{
+							Action:   rbacconfigv3.RBAC_DENY,
+							Policies: map[string]*rbacconfigv3.Policy{},
+						},
+						ShadowRulesStatPrefix: extAuthStatPrefix,
+						Rules: &rbacconfigv3.RBAC{
+							Action:   rbacconfigv3.RBAC_ALLOW,
+							Policies: map[string]*rbacconfigv3.Policy{},
+						},
+					}
+				}
+				extAuthRBAC.ShadowRules.Policies[rule.Name] = rbacPolicy
+				extAuthRBAC.Rules.Policies[rule.Name] = rbacPolicy
+			} else {
+				// InlineTools rule
+				prefixedName := fmt.Sprintf("%s.%s.%s", policy.Namespace, policy.Name, rule.Name)
+				mergedInlineToolsRBAC.Rules.Policies[prefixedName] = rbacPolicy
+				hasInlineToolsRules = true
+			}
+		}
+
+		if hasExtAuthInThisPolicy {
+			// Add common policies to this filter
+			t.addCommonPoliciesToRBAC(extAuthRBAC)
+			extAuthConfigs = append(extAuthConfigs, extAuthRBAC)
+		}
+	}
+
+	// Inject merged InlineTools rules into ExternalAuth filters
+	if len(extAuthConfigs) > 0 && hasInlineToolsRules {
+		for _, extAuthRBAC := range extAuthConfigs {
+			for ruleName, routePolicy := range mergedInlineToolsRBAC.GetRules().GetPolicies() {
+				extAuthRBAC.Rules.Policies[ruleName] = routePolicy
+			}
+		}
+	}
+
+	// Return logic to avoid redundant filters
+	var result []*rbacv3.RBAC
+	if len(extAuthConfigs) > 0 {
+		result = extAuthConfigs
+	} else if hasInlineToolsRules {
+		t.addCommonPoliciesToRBAC(mergedInlineToolsRBAC)
+		result = []*rbacv3.RBAC{mergedInlineToolsRBAC}
+	}
+
+	return result
 }
 
 // buildBackendLevelRBACFilters creates placeholder RBAC filters for backends.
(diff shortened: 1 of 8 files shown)


bowei:
You need to update your commit message to be a well-formatted git commit:

https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

pkg/translator/accesspolicy.go
```diff
@@ -86,6 +87,82 @@ func (t *Translator) buildGatewayLevelRBACFilters(gateway *gatewayv1.Gateway) ([
...
+	var extAuthConfigs []*rbacv3.RBAC
+
+	for _, policy := range policies {
+		var extAuthRBAC *rbacv3.RBAC
+		hasExtAuthInThisPolicy := false
+
+		for _, rule := range policy.Spec.Rules {
+			rbacPolicy, isExternalAuth, extAuthStatPrefix, err := t.translateAccessRuleToRBACPolicy(policy.Namespace, rule)
+			if err != nil {
+				klog.Errorf("Failed to translate access rule %s in policy %s/%s: %v", rule.Name, policy.Namespace, policy.Name, err)
```
bowei, line 109: Seems like we log and then keep going even though there is an error here? Is this correct?

pkg/translator/accesspolicy.go
```diff
@@ -86,6 +87,82 @@ func (t *Translator) buildGatewayLevelRBACFilters(gateway *gatewayv1.Gateway) ([
...
+	hasInlineToolsRules := false
+
+	var extAuthConfigs []*rbacv3.RBAC
+
+	for _, policy := range policies {
+		var extAuthRBAC *rbacv3.RBAC
+		hasExtAuthInThisPolicy := false
+
+		for _, rule := range policy.Spec.Rules {
+			rbacPolicy, isExternalAuth, extAuthStatPrefix, err := t.translateAccessRuleToRBACPolicy(policy.Namespace, rule)
```
bowei, line 107: when you have nested for loops, I would look to extract the innermost nested operation in a separate method that can be tested independently. Can you extract this?

pkg/translator/accesspolicy.go
bowei, line 198: this needs a method to be readable

func refIsXBackend(ref ...) bool {
  return ref.Group != nil && *ref.Group == agenticv0alpha0.GroupName && ref.Kind != nil && *ref.Kind == "XBackend
}

pkg/translator/accesspolicy.go
```diff
@@ -132,14 +209,39 @@ func (t *Translator) calculateMaxBackendRBACFilters(gateway *gatewayv1.Gateway)
 
 	maxCount := 0
 
-	// 3. For each backend, count its accepted policies.
+	// 3. For each backend, count its needed filters after merging.
 	for beKey := range backendNames {
 		policies, err := t.findAccessPoliciesForTarget(agenticv0alpha0.GroupName, "XBackend", beKey.Namespace, beKey.Name)
```
bowei, line 214: this loop inner body should be factored out as a method

pkg/translator/accesspolicy.go
```diff
@@ -132,14 +209,39 @@ func (t *Translator) calculateMaxBackendRBACFilters(gateway *gatewayv1.Gateway)
...
-		count := len(policies)
+
+		hasInlineTools := false
+		extAuthCount := 0
+
+		for _, policy := range policies {
+			hasPolicyExtAuth := false
+
+			for _, rule := range policy.Spec.Rules {
+				if rule.Authorization != nil && rule.Authorization.Type == agenticv0alpha0.AuthorizationRuleTypeExternalAuth {
```
bowei, line 227: this feels like it should be a switch () with case

pkg/translator/accesspolicy.go
```diff
@@ -132,14 +209,39 @@ func (t *Translator) calculateMaxBackendRBACFilters(gateway *gatewayv1.Gateway)
...
+				extAuthCount++
+			}
+		}
+
+		count := extAuthCount
+		if count == 0 && hasInlineTools {
+			count = 1
+		}
+
 		if count > maxCount {
```
bowei, line 245: go 1.25 has max

```
return max(count, max count)
```

pkg/translator/accesspolicy_test.go
```diff
@@ -143,6 +147,122 @@ func TestBuildGatewayLevelRBACFilters(t *testing.T) {
...
+					return p
+				}(),
+			},
+			gatewaysToCheck: map[string][]string{
+				gwName: {fmt.Sprintf("%s%d", constants.GatewayRBACFilterNamePrefix, 1)},
+			},
+			verify: func(t *testing.T, filters []*hcm.HttpFilter) {
+				expectedPrefixedName1 := fmt.Sprintf("%s.%s.%s", ns, "gw-policy-1", "rule-1")
+				expectedPrefixedName2 := fmt.Sprintf("%s.%s.%s", ns, "gw-policy-2", "rule-1")
+				verifyRBACConfig(t, filters, 1, nil, []string{expectedPrefixedName1, expectedPrefixedName2}, 5)
```
haiyanmeng, line 177: Add a comment on `5` to indicate that 3 of them are common policies and two of them are from the accessrules.

pkg/translator/accesspolicy_test.go
```diff
@@ -143,6 +147,122 @@ func TestBuildGatewayLevelRBACFilters(t *testing.T) {
...
+			},
+		},
+		{
+			name: "mixed inline and external auth rules",
```
haiyanmeng, line 181: The evaluation rules for different types are still under review https://github.com/kubernetes-sigs/kube-agentic-networking/pull/220/changes. 

But the parts we are almost certain are:
* [ExternalAuth-type](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/cf8c85b85d067657a5dce7b87270f8099f1e302c/api/v0alpha0/accesspolicy_types.go#L169) Rules are evaluated first before InlineTools-type. If any ExternalAuth rule matching the request returns deny, the request is denied immediately. 
* The community is leaning towards to limit the number of ExternalAuth targeting a given Gateway/XBackend object to 1. 

So I think we will need to create a pair of RBAC http_filter and ext_auth http_filter for each ExternalAuth rule; and these filters need to be evaluated before the RBAC http_filter for the InlineTools rules. 
```
Gateway-level RBAC filter with shadow rules for ExternalAuth
Gateway-level Ext_authz filter
Gateway-level RBAC filter with normal rules for InlineTools
Backend-level RBAC filter with shadow rules for ExternalAuth
Backend-level Ext_authz filter
Backend-level RBAC filter with normal rules for InlineTools
```

`Limiting the number of ExternalAuth targeting a given Gateway/XBackend object to 1`: we can do this in a separate PR if it is easier. To simplify this PR, let us avoid having test coverage for the cases where there are multiple ExternalAuth targeting a given Gateway/XBackend object.

pkg/translator/accesspolicy.go
haiyanmeng, line 1: https://github.com/kubernetes-sigs/kube-agentic-networking/issues/237 says that `change the evaluation rules for multiple InlineTools policies/rules targeting the same Gateway object`. However, the reality is that it is probably more complex to keep the old logic for policies targeting XBackend and new logic for policies targeting Gateway. 

This is to provide more context to you. It is okay to use the new logic for both Gateway-targeted XAccessPolicy and XBackend-targeted XAccessPolicy in this PR. Sorry about the messy situation we are in. Thanks for the understanding!

We most likely would start restricting the target of an XAccessPolicy to Gateway only in v1alpha1. This means we will be able to remove the logic for supporting XBackend-targeted XAccessPolicy in the future.

pkg/translator/accesspolicy_test.go
ziyue-101: I think the current design is that both externalAuth and inlineTool type are associated with a source(SA or SPIFFE). How are they going to work together with same or different sources?

i,e,

1. Are we having logic OR or not for the two rules for the following Xaccesspolicy? 
```
apiVersion: agentic.prototype.x-k8s.io/v0alpha0
kind: XAccessPolicy
metadata:
  name: e2e-backend-combined-policy
spec:
  targetRefs:
    - group: agentic.prototype.x-k8s.io
      kind: XBackend
      name: e2e-mcp-backend
  rules:
    - name: inline-tools-for-tester-sa
      source:
        type: ServiceAccount
        serviceAccount:
          name: e2e-tester-sa
          namespace: e2e-test-ns
      authorization:
        type: InlineTools
        tools:
          - "echo"
          - "get-sum"
    - name: external-auth-for-tester-2-sa
      source:
        type: ServiceAccount
        serviceAccount:
          name: e2e-tester-2-sa
          namespace: e2e-test-ns
      authorization:
        type: ExternalAuth
        externalAuth:
          backendRef:
            kind: Service
            name: authorino
            namespace: e2e-test-ns
            port: 50051
          protocol: GRPC
          grpc: {}
```

2. What is the desired behavior if the second externalauth rule also has the same source as `e2e-tester-sa`

pkg/translator/accesspolicy.go
ziyue-101: Thanks, i will hold off this PR until we have a decision on accesspollices https://github.com/kubernetes-sigs/kube-agentic-networking/pull/220/changes.

I'd like to build the changes incrementally, otherwise part of the change in this part is going to get revered based on the decision outcome.

1. restrict number of externalAuth to 1.
2. re-order the envoy filters.
3. implement logic OR for inLineTools.

ziyue-101:
The PR is targeting v0alpha0 accesspolicy API. 

Now the group has ongoing vaalpah1 api improvement, this PR is no longer relevant.

/close

PR #242, opened 2026-04-17
Complete the support for ReferenceGrant

haiyanmeng:
- Add event handlers for ReferenceGrant in controller to trigger Gateway reconciliation.
- Make AllowedByReferenceGrant in translator generic to accept from/to group and kind.
- Update call sites in backend.go and service.go to use the new signature.
- Implement HTTPRouteBackendRefNamespaceIndex and GatewaySecretRefNamespaceIndex to optimize lookups.
- Restructure enqueueGatewaysForReferenceGrant to use indexers with list fallbacks.
- Replace hardcoded group name strings with gatewayv1.GroupName constant.
- Add comprehensive unit tests for AllowedByReferenceGrant and indexer functions.

**What type of PR is this?**

/kind feature

```release-note
NONE
```

### DIFF
--- a/pkg/controller/referencegrant.go
+++ b/pkg/controller/referencegrant.go
@@ -1,0 +1,181 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"fmt"
+	"reflect"
+
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+	gatewayinformersv1beta1 "sigs.k8s.io/gateway-api/pkg/client/informers/externalversions/apis/v1beta1"
+)
+
+func (c *Controller) setupReferenceGrantEventHandlers(referenceGrantInformer gatewayinformersv1beta1.ReferenceGrantInformer) error {
+	_, err := referenceGrantInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.onReferenceGrantAdd,
+		UpdateFunc: c.onReferenceGrantUpdate,
+		DeleteFunc: c.onReferenceGrantDelete,
+	})
+	return err
+}
+
+func (c *Controller) onReferenceGrantAdd(obj interface{}) {
+	rg := obj.(*gatewayv1beta1.ReferenceGrant)
+	klog.V(4).InfoS("Adding ReferenceGrant", "referencegrant", klog.KObj(rg))
+	c.enqueueGatewaysForReferenceGrant(rg)
+}
+
+func (c *Controller) onReferenceGrantUpdate(old, newObj interface{}) {
+	oldRG := old.(*gatewayv1beta1.ReferenceGrant)
+	newRG := newObj.(*gatewayv1beta1.ReferenceGrant)
+	if !reflect.DeepEqual(oldRG.Spec, newRG.Spec) {
+		klog.V(4).InfoS("Updating ReferenceGrant", "referencegrant", klog.KObj(newRG))
+		c.enqueueGatewaysForReferenceGrant(newRG)
+	}
+}
+
+func (c *Controller) onReferenceGrantDelete(obj interface{}) {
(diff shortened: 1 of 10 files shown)


pkg/controller/referencegrant.go
```diff
@@ -0,0 +1,172 @@
...
+				}
+				if ns != rg.Namespace {
+					continue
+				}
+
+				for _, ref := range route.Spec.ParentRefs {
+					if (ref.Group != nil && string(*ref.Group) != gatewayv1.GroupName) ||
+						(ref.Kind != nil && string(*ref.Kind) != "Gateway") {
+						continue
+					}
```
HarshithaMS005, line 122: @haiyanmeng The ParentRefs filter here duplicates logic that already exists as isGatewayParentRef in pkg/controller/backend.go. That helper encodes the Gateway API defaults for ParentReference: Group and Kind may be omitted and should still be treated as gateway.networking.k8s.io / Gateway.
Using isGatewayParentRef(ref) here keeps behaviour consistent if the rules ever need to change, and avoids two slightly different formulations of the same condition drifting over time.

pkg/controller/referencegrant.go
```diff
@@ -0,0 +1,172 @@
...
+	gatewaysToEnqueue := make(map[string]struct{})
+
+	// 1. Process HTTPRoutes
+	var routeObjs []interface{}
+	if c.gateway.httprouteIndexer != nil {
+		var err error
+		routeObjs, err = c.gateway.httprouteIndexer.ByIndex(HTTPRouteBackendRefNamespaceIndex, rg.Namespace)
+		if err != nil {
+			klog.Errorf("failed to get HTTPRoutes by index: %v", err)
+			return
```
HarshithaMS005: If listing HTTPRoutes by index fails, the function returns before evaluating Gateways that reference Secrets in the ReferenceGrant’s namespace. A grant in namespace N can affect HTTPRoutes with backends in N. A transient or indexing error on one path should not prevent the other path from running.
ByIndex error, log and either fall back to the existing list-based path for that leg only, or set that leg’s object slice to empty and continue, instead of returning from the whole function.

pkg/controller/referencegrant.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	// 2. Process Gateways referencing Secrets
+	var gwObjs []interface{}
+	if c.gateway.gatewayIndexer != nil {
+		var err error
+		gwObjs, err = c.gateway.gatewayIndexer.ByIndex(GatewaySecretRefNamespaceIndex, rg.Namespace)
+		if err != nil {
+			klog.Errorf("failed to get Gateways by index: %v", err)
+			return
```
HarshithaMS005: Same here like in lines 86 - 88 if the Gateway index lookup fails, you never enqueue Gateways discovered via HTTPRoutes. Don’t return before HTTPRoute section completed; still process HTTPRoutes or fall back.

pkg/controller/referencegrant.go
haiyanmeng: Good suggestion. Fixed it.

pkg/controller/referencegrant.go
haiyanmeng: Fixed it

pkg/controller/referencegrant.go
haiyanmeng: Fixed it

pkg/controller/service.go
```diff
@@ -59,6 +59,44 @@ func HTTPRouteServiceRefIndexFunc(obj interface{}) ([]string, error) {
...
+	}
+	var keys []string
+	for _, rule := range route.Spec.Rules {
+		for _, backend := range rule.BackendRefs {
+			backendNS := route.Namespace
+			if backend.Namespace != nil {
+				backendNS = string(*backend.Namespace)
+			}
+			keys = append(keys, backendNS)
+		}
```
HarshithaMS005, line 81: HTTPRouteServiceRefIndexFunc in the same file intentionally skips isXBackendRef(backend.BackendRef) because those backends are not direct Service references. The new HTTPRouteBackendRefNamespaceIndex is used to drive ReferenceGrant-related Gateway reconciliation; including XBackend namespaces here can enqueue Gateways when a ReferenceGrant changes in a namespace that only appears on an XBackend ref, even though Service cross-namespace policy does not apply the same way. Aligning this indexer with the Service ref indexer keeps “which HTTPRoutes matter for Service-style cross-namespace grants” consistent and reduces unnecessary reconciles.

HarshithaMS005:
@haiyanmeng Can you add controller-level tests for enqueueGatewaysForReferenceGrant?

pkg/controller/service.go
haiyanmeng: We do allow an HTTPRoute references an XBackend object, here is an example: https://github.com/kubernetes-sigs/kube-agentic-networking/blob/e0483256f59d0ed894a4957466a7c06ec520af6d/site-src/guides/quickstart/policy/e2e.yaml#L133-L136

Maybe the confusion is from the fact that HTTPRouteBackendRefNamespaceIndex and HTTPRouteBackendRefNamespaceIndexFunc are in `service.go`. Will moving them to `referencegrant.go` help reduce the confusion?

haiyanmeng:
> @haiyanmeng Can you add controller-level tests for enqueueGatewaysForReferenceGrant?

Added.

pkg/controller/service.go
HarshithaMS005: Thanks for the example. Yes moving them would reduce confusion.

haiyanmeng: This PR was merged before I addressed this.

I created https://github.com/kubernetes-sigs/kube-agentic-networking/pull/248 to address this.

PR #244, opened 2026-04-17
Change Envoy proxy Service type to LoadBalancer and add MetalLB to E2E test

haiyanmeng:
Fix translator status reporting and add Hostname support for LoadBalancer
    
    - Fix translator status reporting for listeners and routes.
    - Extend ensureService and Gateway status to handle LoadBalancer Hostnames in addition to IPs.
    - Increased controller retries to 30 to prevent it from giving up on LoadBalancer assignment too early.
    - Increased test-side retries in controller_test.go to 50 times with 10s interval to prevent the test runner from timing out while the controller was still retrying.
    - Update unit tests in proxy_test.go to cover both IP and Hostname cases.
    - Add detailed comments about MetalLB configuration in run-e2e.sh.

**What type of PR is this?**

/kind feature

With the Envoy proxy type `ClusterIP`, the Gateway Address is a cluster-internal IP and not reachable from outside the cluster. 
Gateway API conformance tests are typically executed from the host machine (your laptop or the CI runner), not from within a pod in the cluster.

```release-note
NONE
```

### DIFF
--- a/dev/ci/run-e2e.sh
+++ b/dev/ci/run-e2e.sh
@@ -34,6 +34,54 @@ main() {
   kind delete cluster --name "${CLUSTER_NAME}" || true
   kind create cluster --name "${CLUSTER_NAME}" --config dev/ci/kind-config.yaml --wait 5m
 
+  # MetalLB enables Kind clusters to provide external IPs for LoadBalancer services.
+  # Kind (Kubernetes in Docker) does not come with a native LoadBalancer provider.
+  # Without MetalLB (or a similar tool), any Kubernetes Service of type LoadBalancer (which
+  # is typically used to expose the Gateway) would remain in a <pending> state forever, and
+  # the tests would fail.
+  header "Installing MetalLB"
+  local metallb_version="v0.13.10"
+  kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/"${metallb_version}"/config/manifests/metallb-native.yaml
+
+  needCreate="$(kubectl get secret -n metallb-system memberlist --no-headers --ignore-not-found -o custom-columns=NAME:.metadata.name)"
+  if [ -z "$needCreate" ]; then
+      kubectl create secret generic -n metallb-system memberlist --from-literal=secretkey="$(openssl rand -base64 128)"
+  fi
+
+  # Wait for MetalLB to become available
+  kubectl rollout status -n metallb-system deployment/controller --timeout 5m
+  kubectl rollout status -n metallb-system daemonset/speaker --timeout 5m
+
+  # Configure MetalLB with an IP address pool derived from the Docker network used by Kind.
+  # We inspect the 'kind' network to find the subnet, and take the range .200 to .250.
+  # This range is safe to use because it is at the high end of the subnet, well outside
+  # the range Docker typically uses when dynamically assigning IPs to containers (Kind nodes).
+  # For example, if the subnet is 192.168.8.0/24, address_first_three_octets will
+  # be 192.168.8 and the range will be 192.168.8.200-192.168.8.250.
+  subnet=$(docker network inspect kind | jq -r '.[].IPAM.Config[].Subnet | select(contains(":") | not)')
+  address_first_three_octets=$(echo "${subnet}" | awk -F. '{printf "%s.%s.%s",$1,$2,$3}')
+  address_range="${address_first_three_octets}.200-${address_first_three_octets}.250"
+
+  kubectl apply -f - <<EOF
+apiVersion: metallb.io/v1beta1
+kind: IPAddressPool
+metadata:
+  namespace: metallb-system
+  name: kube-services
+spec:
+  addresses:
+  - ${address_range}
+---
+apiVersion: metallb.io/v1beta1
+kind: L2Advertisement
+metadata:
+  name: kube-services
+  namespace: metallb-system
+spec:
+  ipAddressPools:
+  - kube-services
+EOF
+
   header "Building controller image"
   # These must match the image used in k8s/deploy/deployment.yaml
   REGISTRY="us-central1-docker.pkg.dev/k8s-staging-images/agentic-net"
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -60,7 +60,7 @@ import (
 
 // maxGatewaySyncRetries is the maximum number of rate-limited requeues for a
 // single Gateway key before the worker drops it (avoids infinite hot loops).
-const maxGatewaySyncRetries = 15
+const maxGatewaySyncRetries = 30
 
 // maxBackendFinalizerRetries is the same cap for the XBackend finalizer queue.
 const maxBackendFinalizerRetries = 15
@@ -411,7 +411,7 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
 
 	// Ensure Envoy proxy deployment and service exist.
 	rm := envoy.NewResourceManager(c.core.client, gateway, c.envoyImage, c.agenticIdentityTrustDomain)
-	proxyIP, err := rm.EnsureProxyExist(ctx)
+	proxyAddress, err := rm.EnsureProxyExist(ctx)
 	if err != nil {
 		updateErr := updateGatewayStatus(ctx, c, gateway, newGW, nil, err)
 		return errors.Join(err, updateErr)
@@ -419,15 +419,23 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
 
 	// TODO: Add support for IPv6?
 	newGW.Status.Addresses = []gatewayv1.GatewayStatusAddress{}
-	if net.ParseIP(proxyIP) != nil {
+	if net.ParseIP(proxyAddress) != nil {
+		// proxyAddress is an IP address
 		newGW.Status.Addresses = append(newGW.Status.Addresses,
 			gatewayv1.GatewayStatusAddress{
 				Type:  ptr.To(gatewayv1.IPAddressType),
-				Value: proxyIP,
+				Value: proxyAddress,
+			})
+	} else {
+		// proxyAddress is a hostname
+		newGW.Status.Addresses = append(newGW.Status.Addresses,
+			gatewayv1.GatewayStatusAddress{
+				Type:  ptr.To(gatewayv1.HostnameAddressType),
+				Value: proxyAddress,
 			})
 	}
 
-	logger.Info("Ensured Envoy proxy for gateway exists", "nodeID", rm.NodeID(), "proxyIP", proxyIP)
+	logger.Info("Ensured Envoy proxy for gateway exists", "nodeID", rm.NodeID(), "proxyAddress", proxyAddress)
 
 	// Translate Gateway to xDS resources (includes only current HTTPRoutes/XAccessPolicies; stale rules are omitted).
 	resources, listenerStatuses, httpRouteStatuses, _, err := c.translator.TranslateGatewayToXDS(ctx, gateway)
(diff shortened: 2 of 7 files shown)


pkg/infra/envoy/resourcerender.go
```diff
@@ -304,7 +304,7 @@ func (r *ResourceManager) renderService() *corev1.Service {
 			OwnerReferences: ownerRef(r.gw),
 		},
 		Spec: corev1.ServiceSpec{
-			Type: corev1.ServiceTypeClusterIP,
+			Type: corev1.ServiceTypeLoadBalancer,
```
HarshithaMS005, line 343: @haiyanmeng Setting spec.type to LoadBalancer is correct for exposing the data plane, but the controller still learns the Gateway address from EnsureProxyExist, which today returns svc.Spec.ClusterIP and pkg/controller only publishes an address when that value parses as an IP. For a LoadBalancer Service, clients and conformance tests that read Gateway.status.addresses may therefore still see a cluster-internal IP even though MetalLB/cloud assigns an external address in status.loadBalancer.ingress.

Can you extend this PR (or follow immediately) so that when the proxy Service is LoadBalancer, we wait for status.loadBalancer.ingress and surface that IP or hostname on the Gateway (including hostname type where net.ParseIP does not apply), and tighten waitForServiceReady so we do not proceed only on ClusterIP before the LB assignment exists?

haiyanmeng: Great point. Let me extend this PR to close the gap.

dev/ci/run-e2e.sh
```diff
@@ -34,6 +34,44 @@ main() {
...
+  needCreate="$(kubectl get secret -n metallb-system memberlist --no-headers --ignore-not-found -o custom-columns=NAME:.metadata.name)"
+  if [ -z "$needCreate" ]; then
+      kubectl create secret generic -n metallb-system memberlist --from-literal=secretkey="$(openssl rand -base64 128)"
+  fi
+
+  # Wait for MetalLB to become available
+  kubectl rollout status -n metallb-system deployment/controller --timeout 5m
+  kubectl rollout status -n metallb-system daemonset/speaker --timeout 5m
+
+  # Apply config with addresses based on docker network IPAM
```
bowei: Please add a comment with more detail on how Metal LB is configured and why taking these IP addresses is ok.

dev/ci/run-e2e.sh
```diff
@@ -34,6 +34,44 @@ main() {
   kind delete cluster --name "${CLUSTER_NAME}" || true
   kind create cluster --name "${CLUSTER_NAME}" --config dev/ci/kind-config.yaml --wait 5m
 
+  header "Installing MetalLB"
```
bowei, line 42: Should this be an optional?

haiyanmeng: My local testing + Gemini suggests the answer depends on the environment:

It is required for running these E2E tests in Kind: MetalLB is required by this script. Kind (Kubernetes in Docker) does not come with a native LoadBalancer provider. Without MetalLB (or a similar tool), any Kubernetes Service of type LoadBalancer (which is typically used to expose the Gateway) would remain in a <pending> state forever, and the tests would fail..

It is not required for production or cloud environments: MetalLB is not a hard requirement for the kube-agentic-networking project itself. If you deploy this project in a managed cloud environment (like GKE, AWS, or Azure), the cloud provider automatically provides the LoadBalancer functionality. MetalLB is only needed for bare-metal clusters or local development clusters like Kind that lack a built-in load balancer.

I added a comment to explain why we need to install MetalLB in our e2e test, which runs in a KIND cluster.

dev/ci/run-e2e.sh
haiyanmeng: Done

pkg/infra/envoy/resourcerender.go
haiyanmeng: Done

haiyanmeng:
> @haiyanmeng: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> pull-kube-agentic-networking-e2e	[f8d53ef](https://github.com/kubernetes-sigs/kube-agentic-networking/commit/f8d53ef199a29015ae3ff975203981946788ab3e)	[link](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kube-agentic-networking/244/pull-kube-agentic-networking-e2e/2047049262021218304)	true	`/test pull-kube-agentic-networking-e2e`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes-sigs&repo=kube-agentic-networking&pr=244). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3Ahaiyanmeng). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes-sigs/kube-agentic-networking/issues?q=is:issue+is:open) when you hit one in your PR.
> 
> Details

This test passed locally. I may need to increase the timeout in our e2e test so that it can pass constantly locally and on prow.

LiorLieberman (commented):
Thanks @haiyanmeng ! 
quick question - wouldn't metalLB be required for conformance tests only? i.e when we integrate conformance and we can have something as "run-conformance.sh" or something?

Or a makefile target so we have `make conformance` ? And ideally those will be integrated as well the Gateway-api conformance tests?

haiyanmeng:
> quick question - wouldn't metalLB be required for conformance tests only? i.e when we integrate conformance and we can have something as "run-conformance.sh" or something?

It is required for our e2e test too, which runs in a KIND cluster.

> Or a makefile target so we have make conformance ? And ideally those will be integrated as well the Gateway-api conformance tests?

That is out of the scope for this PR. After closing the gaps in the controller logic, I will send out a PR to add the automatic conformance tests.

pkg/controller/controller.go
```diff
@@ -411,23 +411,31 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
...
 	if err != nil {
 		updateErr := updateGatewayStatus(ctx, c, gateway, newGW, nil, err)
 		return errors.Join(err, updateErr)
 	}
 
 	// TODO: Add support for IPv6?
 	newGW.Status.Addresses = []gatewayv1.GatewayStatusAddress{}
-	if net.ParseIP(proxyIP) != nil {
+	if net.ParseIP(proxyAddress) != nil {
+		// proxyAddress is an IP address
```
bowei: redundant, delete

dev/ci/run-e2e.sh
```diff
@@ -34,6 +34,58 @@ main() {
   kind delete cluster --name "${CLUSTER_NAME}" || true
   kind create cluster --name "${CLUSTER_NAME}" --config dev/ci/kind-config.yaml --wait 5m
 
+  # MetalLB is required for Kind clusters to provide external IPs for LoadBalancer services.
```
LiorLieberman:
```suggestion
  # MetalLB enables Kind clusters to provide external IPs for LoadBalancer services.
```

pkg/controller/controller.go
```diff
@@ -60,7 +60,7 @@ import (
 
 // maxGatewaySyncRetries is the maximum number of rate-limited requeues for a
 // single Gateway key before the worker drops it (avoids infinite hot loops).
-const maxGatewaySyncRetries = 15
+const maxGatewaySyncRetries = 30
```
LiorLieberman, line 63: why needed?

pkg/controller/controller.go
```diff
@@ -411,23 +411,31 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
 
 	// Ensure Envoy proxy deployment and service exist.
 	rm := envoy.NewResourceManager(c.core.client, gateway, c.envoyImage, c.agenticIdentityTrustDomain)
-	proxyIP, err := rm.EnsureProxyExist(ctx)
+	proxyAddress, err := rm.EnsureProxyExist(ctx)
```
LiorLieberman, line 414: nit: whats the motivation to change from proxyIP to proxyAddress

tests/e2e/controller_test.go
```diff
@@ -502,7 +502,8 @@ func deployCommonTestResources(t *testing.T) (namespace, gatewayIP string, clean
 
 	// Return the Gateway IP
 	t.Log("Obtain Gateway Address from status")
-	err = retry(20, 2*time.Second, func() error {
+	// Increased retries to 90 (3 minutes) to give MetalLB and controller enough time to assign and reflect the LoadBalancer IP.
```
LiorLieberman:
```suggestion
```

site-src/guides/quickstart/run-quickstart.sh
```diff
@@ -178,6 +178,9 @@ create_kind_cluster() {
   kubectl config use-context "kind-${CLUSTER_NAME}"
 }
 
+# TODO: install MetalLB after us-central1-docker.pkg.dev/k8s-staging-images/agentic-net/agentic-networking-controller:main
+# includes the changes in https://github.com/kubernetes-sigs/kube-agentic-networking/pull/244
```
LiorLieberman, line 182: is this relevant?

LiorLieberman (commented):
Thanks, mostly LGTM. 

My earlier point mainly asks - are these e2e tests essentially similar to any conformance tests we may write? (applying yaml, expecting a result, asserting result) -- if yes, we may just change these e2e to conformance tests (not the gateway conformance tests, rather the specific conformance tests we will define in this repo in a similar way to Gateway)

Not the scope of this PR - but a good thing to document/figure out the direction

dev/ci/run-e2e.sh
```diff
@@ -34,6 +34,58 @@ main() {
...
+  # For example, if the subnet is 192.168.8.0/24, address_first_three_octets will
+  # be 192.168.8 and the range will be 192.168.8.200-192.168.8.250.
+  subnet=$(docker network inspect kind | jq -r '.[].IPAM.Config[].Subnet | select(contains(":") | not)')
+  address_first_three_octets=$(echo "${subnet}" | awk -F. '{printf "%s.%s.%s",$1,$2,$3}')
+  address_range="${address_first_three_octets}.200-${address_first_three_octets}.250"
+
+  # Apply MetalLB configuration:
+  # 1. IPAddressPool: Defines the range of IPs that MetalLB can assign to LoadBalancer services.
+  # 2. L2Advertisement: Instructs MetalLB to use Layer 2 mode (ARP) to advertise these IPs,
+  #    which is suitable for Kind clusters where BGP is not available.
```
LiorLieberman:
```suggestion
```
Nit: too verbose. we dont need it

haiyanmeng:
> My earlier point mainly asks - are these e2e tests essentially similar to any conformance tests we may write? (applying yaml, expecting a result, asserting result) -- if yes, we may just change these e2e to conformance tests (not the gateway conformance tests, rather the specific conformance tests we will define in this repo in a similar way to Gateway)

This PR does not add new e2e tests. I don't plan to add extra test coverage which are already covered by the Gateway API conformance tests.

dev/ci/run-e2e.sh
haiyanmeng: Done

dev/ci/run-e2e.sh
haiyanmeng: Done

pkg/controller/controller.go
haiyanmeng: 15 is not enough sometimes.

pkg/controller/controller.go
haiyanmeng: proxy address could be IP address or hostname.

site-src/guides/quickstart/run-quickstart.sh
haiyanmeng: Yes. We need to update our quickstart after this change is merged.

LiorLieberman (commented):
/lgtm

PR #246, opened 2026-04-20
envoy: server-side apply, skip unchanged applies, and ConfigMap checksum for pod rollouts

### LINKED ISSUE #238: [Improvement] Reconcile Envoy Proxy Infrastructure Changes
## Description
The current implementation of the `ResourceManager` in `pkg/infra/envoy/proxy.go` only ensures the *existence* of Envoy proxy infrastructure resources (Deployment, Service, ConfigMap, ServiceAccount). It does not reconcile or update these resources if they already exist but their desired state has changed.
While traffic routing and policies are updated dynamically via the xDS server, any changes to the proxy's infrastructure are currently ignored after the initial creation.
## Current Behavior
In `pkg/infra/envoy/proxy.go`, the `EnsureProxyExist` method and its helpers (like `ensureDeployment`, `ensureConfigMap`) follow a "create-if-missing" pattern:
1. They perform a `Get` to check if the resource exists.
2. If not found, they `Create` it.
3. If it exists, they do nothing, even if the desired spec (e.g., image version or bootstrap config) has changed.
## Why This Matters
Without infrastructure reconciliation, the controller cannot support:
*   **Upgrades**: Changing the `envoyImage` passed to the controller will not update existing Envoy deployments.
*   **Config Drift**: Manual modifications to the Envoy deployments or services will not be reverted by the controller.
*   **Bootstrap Updates**: Changes to the rendered ConfigMap (bootstrap config) are not applied to existing instances.
## Expected Behavior
The controller should reconcile the infrastructure resources to match the desired state rendered by the `ResourceManager`. 
*   If the desired Deployment or ConfigMap differs from the current state in the cluster, the controller should update the resource.
*   Appropriate rollout strategies should be considered (e.g., rolling update for the Deployment).
## Possible Implementation Approach
Instead of just checking for existence, we can use an `Update` or `Patch` strategy:
*   For the **Deployment**, we could check if the spec has changed (e.g., image or labels) and call `Update`.
*   For the **ConfigMap**, we could update the data if it differs.

HarshithaMS005:
**What type of PR is this?**
/kind feature

**What this PR does **:
The Envoy ResourceManager used to only create the proxy ServiceAccount, ConfigMap, Deployment, and Service when they were missing. This PR reconciles those objects on every sync so changes (Envoy image, bootstrap/SDS content, manual edits) converge without deleting the Gateway.
Infra updates use server-side apply ([docs](https://kubernetes.io/docs/reference/using-api/server-side-apply/)) with a single field manager and Force, instead of field-by-field comparison in reconcile.go. reconcile.go now only holds configMapDataChecksum for the pod-template annotation. Because updating a mounted ConfigMap alone does not restart pods, the Deployment pod template keeps a checksum annotation derived from rendered ConfigMap data so bootstrap/SDS changes still trigger a rollout. The controller does not wait for the Envoy Deployment Available condition, so it does not block reconciliation of other Gateways ([#198](https://github.com/kubernetes-sigs/kube-agentic-networking/pull/198)).
Proxy resources use the standard Gateway label gateway.networking.k8s.io/gateway-name ([GEP-1762](https://gateway-api.sigs.k8s.io/geps/gep-1762/#automated-deployments)).

Fixes #238 

```release-note
NONE
```

### DIFF
--- a/pkg/infra/envoy/proxy.go
+++ b/pkg/infra/envoy/proxy.go
@@ -22,8 +22,8 @@ import (
 	"encoding/hex"
 	"fmt"
 
-	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -68,8 +68,8 @@ func proxyName(namespace, name string) string {
 	return fmt.Sprintf(constants.ProxyNameFormat, hex.EncodeToString(hash[:6]))
 }
 
-// EnsureProxyExist ensures that the Envoy proxy deployment, service, and other resources exist and are ready.
-// It returns the LoadBalancer address (IP or Hostname) of the proxy service.
+// EnsureProxyExist ensures that the Envoy proxy deployment, service, and other resources exist and match desired state.
+// It returns the LoadBalancer address (IP or Hostname) of the proxy service when assigned.
 func (r *ResourceManager) EnsureProxyExist(ctx context.Context) (string, error) {
 	logger := klog.FromContext(ctx).WithValues("resourceName", klog.KRef(r.namespace, r.nodeID))
 	ctx = klog.NewContext(ctx, logger)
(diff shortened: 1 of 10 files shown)


pkg/infra/envoy/reconcile.go
```diff
@@ -0,0 +1,213 @@
+/*
+Copyright 2025 The Kubernetes Authors.
```
haiyanmeng: Remove 2025

pkg/infra/envoy/reconcile_test.go
```diff
@@ -0,0 +1,242 @@
+/*
+Copyright 2025 The Kubernetes Authors.
```
haiyanmeng: Remove 2025

pkg/infra/envoy/proxy.go
haiyanmeng, line 1: Let us use Server Side Apply https://kubernetes.io/docs/reference/using-api/server-side-apply/ for updating envoy infra resources. This would eliminate the filed comparison logic in reconcile.go

pkg/infra/envoy/proxy.go
```diff
@@ -143,61 +161,81 @@ func (r *ResourceManager) ensureConfigMap(ctx context.Context) error {
...
+		if syncDeployment(merged, want) {
+			if _, err = r.client.AppsV1().Deployments(want.Namespace).Update(ctx, merged, metav1.UpdateOptions{}); err != nil {
+				return fmt.Errorf("failed to update envoy deployment: %w", err)
+			}
+			logger.Info("Updated envoy proxy deployment to match desired state (rolling update proceeds per Deployment strategy)")
 		}
 	}
 
-	return fmt.Errorf("envoy deployment %s is not available yet", deployment.Name)
+	if err := waitForDeploymentAvailable(ctx, r.client, want.Namespace, want.Name); err != nil {
```
haiyanmeng: We deliberately removed waitForDeploymentAvailable in https://github.com/kubernetes-sigs/kube-agentic-networking/pull/198 to avoid the situation where the agentic networking controller is stuck at waiting for a Envoy Deployment to become available and cannot reconcile other Gateway resources.

pkg/infra/envoy/proxy.go
HarshithaMS005: Sure.

pkg/infra/envoy/proxy.go
HarshithaMS005: Okay i will work on it.

bowei:
I think since we are adding this new server-side apply logic, it would be good to send a PR for one resource for review to avoid having a quite large review. Once that review is done, we can have a follow up for the rest of the resources as they will be boilerplate in a single follow-up PR.

In addition, it would be good to discuss explicitly which fields you chose to be managed vs unmanaged. There are quite a lot of choices here and it's not clear that the choices are obvious?

Finally, the checksum on the ConfigMap. I would like this separate out into its own PR.

pkg/infra/envoy/proxy.go
```diff
@@ -93,111 +92,74 @@ func (r *ResourceManager) NodeID() string {
...
-			if err != nil {
-				return fmt.Errorf("failed to create envoy serviceaccount: %w", err)
-			}
-		} else {
-			return fmt.Errorf("failed to get envoy serviceaccount: %w", err)
-		}
-	}
-	logger.Info("Envoy proxy serviceaccount is ready!")
+	want := r.renderServiceAccount()
+	if _, err := r.client.CoreV1().ServiceAccounts(want.Namespace).Apply(ctx, serviceAccountApply(want), envoyInfraApplyOptions()); err != nil {
```
haiyanmeng, line 114: Can we check the existence of the object before applying and only apply the object whent there are diffs?

HarshithaMS005: Sure

HarshithaMS005:
@bowei , Thanks for the review — agree on all three points.

- In this branch i ended up with stacked commits (ServiceAccount skip first, then ConfigMap / Deployment / Service skip) rather than separate GitHub PRs; happy to mirror that in the PR description or split into follow-up PRs if you’d rather merge SA-only first.
- Managed vs unmanaged: Documented in serviceAccountDesiredMatchesExisting and in proxy_desired_match.go for ConfigMap, Deployment, and Service (configMapDesiredMatchesExisting, deploymentDesiredMatchesExisting, serviceDesiredMatchesExisting), aligned with what each serviceAccountApply/configMapApply/deploymentApply/
serviceApply sends on SSA.
- Checksum/rollout: Understood on isolating that story. Today it ships with the SSA infra work on this branch; we can either call that out explicitly in the description/release notes, or I can follow up with a small PR that only moves/splits checksum-related files/commits if you want a distinct review for “hash ConfigMap data → pod rollout.” Say what you prefer.

pkg/infra/envoy/proxy.go
```diff
@@ -101,116 +101,157 @@ func (r *ResourceManager) NodeID() string {
 	return r.nodeID
 }
 
-// ensureSA ensures that the ServiceAccount for the Envoy proxy exists.
+// ensureSA applies the ServiceAccount for the Envoy proxy (server-side apply).
```
bowei: There are some resources we should just consider "owned completely" by the controller. In these cases, it's not out of the question to do the simple thing and overwrite all fields. Is Service Account one of these cases?

I'm trying to think through which controllers/users will "collaborate" on the definition of the service account.

HarshithaMS005: Yes — the Envoy proxy ServiceAccount is fully owned by this controller. It has a deterministic hashed name, a Gateway ownerReference, is only referenced by the proxy Deployment, and is deleted with the Gateway. We don’t expect users or other controllers to co-manage this object directly; the only supported customization path is Gateway.spec.infrastructure labels/annotations propagated to proxy infra.
We still shouldn’t do a full-object replace that clears .secrets (populated by the token controllers). SSA via serviceAccountApply only sets labels, annotations, and ownerReferences with our field manager + Force, which is the simple “overwrite what we own” model without touching apiserver-owned fields.

I removed the GET/compare skip-apply logic for the ServiceAccount — that optimization made sense for shared/complex resources but was unnecessary here. ensureSA now always Applies. ConfigMap/Deployment/Service still use skip-apply where the field surface is larger.

PR #249, opened 2026-04-22
Fix translator status reporting for listeners and routes

haiyanmeng:
- Use routesByListener map length for AttachedRoutes instead of error-prone counter.

- Set Programmed condition to False on listeners with unresolved references.

- Populate default fields for resolvedRefsCondition in validateHTTPRoute.

- Add unit tests for AttachedRoutes and unresolved references in translator_test.go.

**What type of PR is this?**

/kind cleanup

```release-note
NONE
```

### DIFF
--- a/pkg/translator/translator.go
+++ b/pkg/translator/translator.go
@@ -186,7 +186,6 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
 
 		// All these listeners have the same port
 		for _, listener := range listeners {
-			var attachedRoutes int32
 			listenerStatus := gatewayv1.ListenerStatus{
 				Name:           listener.Name,
 				SupportedKinds: []gatewayv1.RouteGroupKind{},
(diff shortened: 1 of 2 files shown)


pkg/translator/translator.go
```diff
@@ -308,7 +315,8 @@ func (t *Translator) buildEnvoyResourcesForGateway(gateway *gatewayv1.Gateway) (
 				filterChains = append(filterChains, filterChain)
 			}
 
-			listenerStatus.AttachedRoutes = attachedRoutes
+			//nolint:gosec // G115: Number of routes on a listener is well within int32 range.
```
LiorLieberman, line 318: what does G115 mean?

LiorLieberman (approved):
Thanks @haiyanmeng, one nit on the comment but otherwise LGTM

pkg/translator/translator.go
haiyanmeng: https://github.com/securego/gosec/blob/553d8a50502028375f270b69e959dc11c7952608/RULES.md?plain=1#L39

PR #250, opened 2026-04-22
Validate cert refs in gateway listener and add test coverage

haiyanmeng:
- Implemented proper validation for CertificateRefs in TLS listeners, checking for secret existence and verifying cross-namespace references against ReferenceGrant.

- Added unit tests in listener_test.go covering various valid and invalid scenarios for listener references.

**What type of PR is this?**

/kind cleanup

```release-note
NONE
```

### DIFF
--- a/pkg/translator/listener.go
+++ b/pkg/translator/listener.go
@@ -58,30 +58,28 @@ const (
 	wellknownJWTAuthnFilter = "envoy.filters.http.jwt_authn"
 )
 
-// setListenerCondition is a helper to safely set a condition on a listener's status
-// in a map of conditions.
-func setListenerCondition(
-	conditionsMap map[gatewayv1.SectionName][]metav1.Condition,
-	listenerName gatewayv1.SectionName,
-	condition metav1.Condition,
-) {
-	// This "get, modify, set" pattern is the standard way to
-	// work around the Go constraint that map values are not addressable.
-	conditions := conditionsMap[listenerName]
+type listenerConditions map[gatewayv1.SectionName][]metav1.Condition
+
+func (lc listenerConditions) setCondition(listenerName gatewayv1.SectionName, condition metav1.Condition) {
+	conditions := lc[listenerName]
 	if conditions == nil {
 		conditions = []metav1.Condition{}
 	}
 	meta.SetStatusCondition(&conditions, condition)
-	conditionsMap[listenerName] = conditions
+	lc[listenerName] = conditions
+}
+
+func (lc listenerConditions) isConditionTrue(listenerName gatewayv1.SectionName, conditionType string) bool {
+	return meta.IsStatusConditionTrue(lc[listenerName], conditionType)
 }
 
 // validateListeners checks for conflicts among all listeners on a Gateway as per the spec.
 // It returns a map of conflicted listener conditions and a Gateway-level condition if any conflicts exist.
-func (t *Translator) validateListeners(gateway *gatewayv1.Gateway) map[gatewayv1.SectionName][]metav1.Condition {
-	listenerConditions := make(map[gatewayv1.SectionName][]metav1.Condition)
+func (t *Translator) validateListeners(gateway *gatewayv1.Gateway) listenerConditions {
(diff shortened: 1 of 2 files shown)


pkg/translator/listener.go
bowei: extract listenerConditions as a type with helper methods:

suggest

```
type listenerConditions map[gatewayv1.SectionName][]metav1.Condition

func (lc listenerConditions) setCondition(...) { ... }

func (lc listenerConditions) isConditionTrue(...) bool { ... }
```

then the code below will be easier to understand

haiyanmeng: Done

ziyue-101:
Hi, haiyan, do you plan to merge #231 or this PR first. 

I think this PR would be a nice followup for #231. I am writhing the logic in #231 to default to gateway namespace for reference certs(which we can support cross-namespace reference in a followup)

pkg/translator/listener.go
```diff
@@ -142,20 +141,94 @@ func (t *Translator) validateListeners(gateway *gatewayv1.Gateway) map[gatewayv1
 
 	for _, listener := range gateway.Spec.Listeners {
 		// If a listener is already conflicted, we don't need to check its secrets.
-		if meta.IsStatusConditionTrue(listenerConditions[listener.Name], string(gatewayv1.ListenerConditionConflicted)) {
```
LiorLieberman, line 145: nit: would rather leave `	return meta.IsStatusConditionTrue(lc[listenerName], conditionType)` and not have another function to wrap it (avoiding another human hop to read another function)

pkg/translator/listener.go
```diff
@@ -57,30 +58,28 @@ const (
...
-// in a map of conditions.
-func setListenerCondition(
-	conditionsMap map[gatewayv1.SectionName][]metav1.Condition,
-	listenerName gatewayv1.SectionName,
-	condition metav1.Condition,
-) {
-	// This "get, modify, set" pattern is the standard way to
-	// work around the Go constraint that map values are not addressable.
-	conditions := conditionsMap[listenerName]
+type listenerConditions map[gatewayv1.SectionName][]metav1.Condition
```
LiorLieberman, line 61: making this a type is a little confusing to me (enforcing the reader to reason about another type that indirectly is a map of conditions). Can we leave the type out, and leave the function below as a func and not a method.

haiyanmeng: Making this a type is a suggestion from @bowei : https://github.com/kubernetes-sigs/kube-agentic-networking/pull/250#discussion_r3127414500

LiorLieberman: I see. I dont love that but can revisit in the future. Thanks for the context

pkg/translator/listener.go
haiyanmeng: @LiorLieberman  I am not following your comment. Can you clarify it?

haiyanmeng:
> Hi, haiyan, do you plan to merge #231 or this PR first.
> 
> I think this PR would be a nice followup for #231. I am writhing the logic in #231 to default to gateway namespace for reference certs(which we can support cross-namespace reference in a followup)

@ziyue-101 , this PR is pretty straightforward and I would like to get it merge ASAP. 
I notice that https://github.com/kubernetes-sigs/kube-agentic-networking/pull/231 is in draft mode currently. Let me know if you have any concerns over this.

pkg/translator/listener.go
LiorLieberman: the code introduces 
```
func (lc listenerConditions) isConditionTrue(listenerName gatewayv1.SectionName, conditionType string) bool {
	return meta.IsStatusConditionTrue(lc[listenerName], conditionType)
}
```

I am saying that its better to just leave `meta.IsStatusConditionTrue` -- no need to wrap it in another function. I think reading meta.IsStatusConditionTrue is more readable than `<someStruct>.IsConditionTrue`

haiyanmeng: Thanks for the clarification.

LiorLieberman (commented):
/lgtm

PR #251, opened 2026-04-23
Install MetalLB in quickstart and fix envoy proxy label

haiyanmeng:
- Move MetalLB installation logic from dev/ci/run-e2e.sh to dev/ci/lib.sh, and update both run-e2e.sh and run-quickstart.sh to use it.
- Use the correct label to check the existence/readiness of envoy deployment and SA.

The reason for installing MetalLB in quickstart is that https://github.com/kubernetes-sigs/kube-agentic-networking/pull/244 switched the Envoy Proxy Service type from `ClusterIP` to `LoadBalancer`.

**What type of PR is this?**

/kind bug

```release-note
NONE
```

### DIFF
--- a/dev/ci/lib.sh
+++ b/dev/ci/lib.sh
@@ -1,0 +1,74 @@
+#!/usr/bin/env bash
+
+# Copyright The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# Install MetalLB on a kind cluster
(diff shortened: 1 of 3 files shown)


dev/ci/lib.sh
```diff
@@ -0,0 +1,68 @@
...
+  kubectl rollout status -n metallb-system deployment/controller --timeout 5m
+  kubectl rollout status -n metallb-system daemonset/speaker --timeout 5m
+
+  # Configure MetalLB with an IP address pool derived from the Docker network used by Kind.
+  # We inspect the 'kind' network to find the subnet, and take the range .200 to .250.
+  # This range is safe to use because it is at the high end of the subnet, well outside
+  # the range Docker typically uses when dynamically assigning IPs to containers (Kind nodes).
+  # For example, if the subnet is 192.168.8.0/24, address_first_three_octets will
+  # be 192.168.8 and the range will be 192.168.8.200-192.168.8.250.
+  subnet=$(docker network inspect kind | jq -r '.[].IPAM.Config[].Subnet | select(contains(":") | not)')
```
guicassolato: This only works with Docker. A Podman-compatible equivalent would probably be:

```
podman network inspect kind | jq -r '.[].subnets.[].subnet | select(contains(":") | not)'
```

Could we maybe introduce a `CONTAINER_ENGINE` env var to bend the behaviour accordingly?

haiyanmeng: Good suggestion. Done

guicassolato (approved):
/lgtm
/approve

Issue #255, opened 2026-04-23
Release 0.1


LiorLieberman:
Opening an issue to track the first 0.1 experimental release. Scope for this release includes only XAccessPolicy api.

david-martin:
https://github.com/kubernetes-sigs/kube-agentic-networking/releases/tag/v0.1.0

PR #259, opened 2026-04-29
Add v1alpha1 api

LiorLieberman:
**What type of PR is this?**

Add v1alpha1 API
Move ExternalAuth to top level and add more native mcp support as discussed.

Relates to #255 

```release-note
add v1alpha1 api for XAccessPolicy
```


haiyanmeng (commented):
@LiorLieberman , can you add test coverage into https://github.com/kubernetes-sigs/kube-agentic-networking/tree/main/tests/cel and https://github.com/kubernetes-sigs/kube-agentic-networking/tree/main/tests/crd? Thanks!

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+	//
+	// * Gateway
+	//
+	// This API may be extended in the future to support additional kinds of targetRefs.
+	// Implementations may support additional kinds in an implementation specific manner.
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=10
+	// +listType=atomic
+	// +kubebuilder:validation:XValidation:rule="self.all(ref, ref.kind == self[0].kind)",message="All targetRefs must have the same Kind"
```
haiyanmeng, line 45: why https://github.com/kubernetes-sigs/kube-agentic-networking/blob/9d6bbc10e13f7a30040ab48e45344ff387d6f6d8/api/v0alpha0/accesspolicy_types.go#L35 is not included in this PR?

If we decide to keep supporting Backend-targeted AccessPolicy in the reference implementation, we should allow the target to be either Gatway or Backend.

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+	// Evaluation logic:
+	// 1. ExternalAuth runs before all other Allow policies.
+	// 2. If an ExternalAuth server denies the request, the request is denied.
+	// 3. If it allows the request, processing continues for all other allow policies for that target.
+	// 4. The request is allowed only if all allow policies allow it.
+	// +required
+	Action ActionType `json:"action"`
+
+	// ExternalAuth specifies an external auth filter to be used for authorization.
+	// Core support is limited to 1 extAuth callout per target.
```
haiyanmeng: extAuth -> externalAuth

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+	// +listType=atomic
+	// +kubebuilder:validation:XValidation:rule="self.all(ref, ref.kind == self[0].kind)",message="All targetRefs must have the same Kind"
+	TargetRefs []gwapiv1.LocalPolicyTargetReferenceWithSectionName `json:"targetRefs"`
+
+	// Action specifies the action to be taken when rules match.
+	// Evaluation logic:
+	// 1. ExternalAuth runs before all other Allow policies.
+	// 2. If an ExternalAuth server denies the request, the request is denied.
+	// 3. If it allows the request, processing continues for all other allow policies for that target.
+	// 4. The request is allowed only if all allow policies allow it.
```
haiyanmeng, line 53: https://github.com/kubernetes-sigs/kube-agentic-networking/pull/220 proposes that the request is allowed if any allow policy allows it.

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+	// +required
+	Action ActionType `json:"action"`
```
haiyanmeng: Should we have a default value for `action`?

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+	// Authorization specifies the authorization rule to be applied to requests from the source.
+	// +optional
+	Authorization *AuthorizationRule `json:"authorization,omitempty"`
+}
+
+// Source specifies the source of a request.
+//
+// Type must be set to indicate the type of source type.
+// Similarly, either SPIFFE or Serviceaccount can be set based on the type.
+type Source struct {
```
haiyanmeng: Please incorporate the changes proposed in https://github.com/kubernetes-sigs/kube-agentic-networking/discussions/235#discussioncomment-16519995 in this PR.

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=63
+	Name string `json:"name"`
+	// Source specifies the source of the request.
+	// +required
+	Source Source `json:"source"`
+	// Authorization specifies the authorization rule to be applied to requests from the source.
+	// +optional
+	Authorization *AuthorizationRule `json:"authorization,omitempty"`
```
haiyanmeng, line 100: https://github.com/kubernetes-sigs/kube-agentic-networking/discussions/235#discussioncomment-16519946 proposed move the fields under `authorization` directly into `AccessRule`. What is the rationale to keep it?

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+// +kubebuilder:validation:Enum=tools;prompts;resources;prompts/list;tools/list;resources/list;resources/templates/list;prompts/get;tools/call;resources/subscribe;resources/unsubscribe;resources/read
+type MCPMethodName string
+
+// AuthorizationRuleType identifies a type of authorization rule.
+// +kubebuilder:validation:Enum=Inline
+type AuthorizationRuleType string
+
+const (
+	// AuthorizationRuleTypeInline is used to identify authorization rules
+	// declared as attributes inside the policy (inline)
```
haiyanmeng, line 231: what does `policy` mean?

why is `(inline)` needed here?

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+	AuthorizationRuleTypeInline AuthorizationRuleType = "Inline"
+
+	// TODO: Add CEL support
```
haiyanmeng: Remove this TODO for now

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+type AuthorizationRule struct {
+	// +unionDiscriminator
+	// +required
+	Type AuthorizationRuleType `json:"type"`
+
+	// MCP defines MCP-specific matching criteria.
+	// +optional
+	MCP MCPAttributes `json:"mcp,omitempty"`
```
haiyanmeng, line 179: what does an AuthorizationRule without an `mcp` field do?

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+// MCPAttributes defines the protocol-specific attributes for MCP authorization.
+type MCPAttributes struct {
+	// Methods is a list of specific MCP functional methods to match.
+	// +kubebuilder:validation:MaxItems=10
```
haiyanmeng, line 189: Should we require minItems to be 1?

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+	// Allowed values:
+	// 1. 'tools', 'prompts', 'resources': Matches all sub-methods under these categories.
+	// 2. 'prompts/list', 'tools/list', 'resources/list', 'resources/templates/list'.
+	// 3. 'prompts/get', 'tools/call', 'resources/subscribe', 'resources/unsubscribe', 'resources/read'.
+	// Parameters cannot be specified for categories 1 and 2.
+	// +required
+	Name MCPMethodName `json:"name"`
+
+	// Params allows matching against specific arguments in the MCP request.
+	// Only valid for 'get', 'call', 'subscribe', 'unsubscribe', and 'read' methods.
```
haiyanmeng, line 209: Use a kubebuilder directive for this

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+// MCPMethod defines a specific MCP method and its associated parameters.
+type MCPMethod struct {
+	// Name is the MCP method to match against (e.g., 'tools/call').
+	// Allowed values:
+	// 1. 'tools', 'prompts', 'resources': Matches all sub-methods under these categories.
+	// 2. 'prompts/list', 'tools/list', 'resources/list', 'resources/templates/list'.
+	// 3. 'prompts/get', 'tools/call', 'resources/subscribe', 'resources/unsubscribe', 'resources/read'.
+	// Parameters cannot be specified for categories 1 and 2.
```
haiyanmeng, line 204: Use a kubebuilder directive for this

haiyanmeng, line 204: This seems inconsistent with line 199

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+
+	// Params allows matching against specific arguments in the MCP request.
+	// Only valid for 'get', 'call', 'subscribe', 'unsubscribe', and 'read' methods.
+	// +optional
+	//+listType=set
+	// +kubebuilder:validation:MaxItems=10
+	Params []MCPMethodParam `json:"params,omitempty"`
+}
+
+// +kubebuilder:validation:MaxLength=20
```
haiyanmeng, line 218: 20 seems pretty limited?

haiyanmeng:
> pull-kube-agentic-networking-e2e

@LiorLieberman , this seems unexpected. PTAL.

api/v1alpha1/accesspolicy_types.go
LiorLieberman: Good question. I think it should be required and not have a default value, so we encourage a more explicit approach

api/v1alpha1/accesspolicy_types.go
LiorLieberman: I think its better to start small and relax validation if needed. I dont think that someone will need a parameter that is more than 20 chars? do you have anything in mind that suggests params are more than 20 chars long?

api/v1alpha1/accesspolicy_types.go
LiorLieberman: Hi Haiyan, why is this inconsistent? 

191-193 are the methods that can be specified. 

Separately 194, is saying that only for `'prompts/get', 'tools/call', 'resources/subscribe', 'resources/unsubscribe', 'resources/read'` you can set params. The reason is that the methods decsribed in #1 and #2 dont really have the params. 

199 repeats 194.

api/v1alpha1/accesspolicy_types.go
bowei: I think given there is no official support for XBackend, we should leave references to it out of the validation. When the support is official, then it makes sense to add it back in.

LiorLieberman: Hi Haiyan, as we indicated before, the release is not going to include XBackend in it. Therefore Core support for targeting AccessPolicy is going to be only Gateway. However, we still want to allow room for extensibility for targeting any GVK (one example would be the protoype in this repo which will continue to target v0alpha0 backend) so can not have this validation.

api/v1alpha1/accesspolicy_types.go
LiorLieberman: Resolving, feel free to reopen if you have any more concerns

api/v1alpha1/accesspolicy_types.go
LiorLieberman: good question, to avoid more changes to the api that have not been discussed, I think we can follow up on that on the next release. Personally, I am definitely interested in discussing it and consider moving the fields under AccessRule.

api/v1alpha1/accesspolicy_types.go
LiorLieberman: Policy mean the CRD. let me add a clarification. 
Inline means the rules are defined in the policy vs CEL which is a more dynamic expression that can allow variety of things.

LiorLieberman: I think we can discuss on naming if this is confusing

api/v1alpha1/accesspolicy_types.go
LiorLieberman: Added a clarification.

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+	// +optional
+	ExternalAuth *gwapiv1.HTTPExternalAuthFilter `json:"externalAuth,omitempty"`
+
+	// Rules defines a list of rules to be applied to the target.
+	// An AccessPolicy must have at least one rule.
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=10
+	// +listType=atomic
+	// +kubebuilder:validation:XValidation:rule="self.all(r, self.filter(x, x.name == r.name).size() == 1)",message="AccessRule names must be unique"
```
bowei, line 68: [minor] isn't this a n^2 validation? The limit right now is 10 so it probably is ok

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+	// ActionTypeExternalAuth is used to identify that the request should be delegated to an external auth service if rules match.
+	ActionTypeExternalAuth ActionType = "ExternalAuth"
+)
+
+// AccessRule specifies an authorization rule for the targeted backend.
+// If the tool list is empty, the rule denies access to all tools from Source.
+type AccessRule struct {
+	// Name specifies the name of the rule.
+	// +required
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
```
bowei, line 90: you should say it follows DNS label rule

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,294 @@
...
+	// If not specified, current namespace (the namespace of the policy) is used.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+
+	// Name is the name of the ServiceAccount.
+	// +required
+	Name string `json:"name"`
+}
+
+type AuthorizationRule struct {
```
bowei, line 170: docstring

api/v1alpha1/accesspolicy_types.go
LiorLieberman: While we can let ExternalAuth **deny** early. Having it allow early has several concerns.

External Auth servers typically handle global context (like JWTs or tokens), but they don't necessarily  anything about local boundaries inside the cluster. There may be cases where we still need our downstream policies (like SPIFFE identity matching or MCP method whitelisting) to validate that workload A is actually authorized to communicate with workload B, **after external auth callout.**

Hope thats clear

bowei: Looks like that proposal is still in discussion? v1alpha1 is getting cut with the current behavior -- it is still alpha and can change.

api/v1alpha1/accesspolicy_types.go
LiorLieberman: yeah good callout. The 10 limit making it ok.

api/v1alpha1/accesspolicy_types.go
LiorLieberman: this is already validated on `MCPMethodName` see

```
// MCPMethodName defines the allowed MCP methods for matching.
// +kubebuilder:validation:Enum=tools;prompts;resources;prompts/list;tools/list;resources/list;resources/templates/list;prompts/get;tools/call;resources/subscribe;resources/unsubscribe;resources/read
type MCPMethodName string
```

api/v1alpha1/accesspolicy_types.go
LiorLieberman: done.

api/v1alpha1/accesspolicy_types.go
LiorLieberman: I added 

```
	// If empty or omitted, no method-level whitelisting is applied, meaning all
	// MCP methods (e.g., all tools, prompts, and resources) are permitted.
	// +kubebuilder:validation:MaxItems=10
```

LiorLieberman:
> @LiorLieberman: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> pull-kube-agentic-networking-e2e	[dee5b7d](https://github.com/kubernetes-sigs/kube-agentic-networking/commit/dee5b7d8c33deb89ec93cb32b3069cb25b1f7181)	[link](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kube-agentic-networking/259/pull-kube-agentic-networking-e2e/2050352897719275520)	true	`/test pull-kube-agentic-networking-e2e`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes-sigs&repo=kube-agentic-networking&pr=259). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3ALiorLieberman). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes-sigs/kube-agentic-networking/issues?q=is:issue+is:open) when you hit one in your PR.
> 
> Details

This test is very flaky. This API should not cause this to fail. 

I think its is related to @ziyue-101 's last changes with TLS.

bowei:
/lgtm

But you need to get the tests to pass

api/v1alpha1/accesspolicy_types.go
haiyanmeng: Should we validate the targetRef in control plane?

api/v1alpha1/accesspolicy_types.go
haiyanmeng: > While we can let ExternalAuth deny early. Having it allow early has several concerns.

This is not what #220  proposes. `allow` policies means `inline` policies. I am trying to suggest that we define the rules here to address https://github.com/kubernetes-sigs/kube-agentic-networking/issues/237.

api/v0alpha0/accesspolicy_types.go
```diff
@@ -205,7 +205,6 @@ type AccessPolicyStatus struct {
 // +genclient
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-
```
haiyanmeng, line 208: undo this chnage

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,303 @@
...
+
+// AccessRule specifies an authorization rule for a specified target.
+type AccessRule struct {
+	// Name specifies the name of the rule.
+	// This follows the DNS Subdomain naming convention.
+	// See: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
+	// +required
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
```
haiyanmeng: In v0alpha0, maxLength is set to 63: https://github.com/kubernetes-sigs/kube-agentic-networking/blob/c1f642f7fcf39a080e418a83fda584fd3d0f90fa/api/v0alpha0/accesspolicy_types.go#L56

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,303 @@
...
+	// If omitted, all access from the specified source is allowed.
+	// +optional
+	Authorization *AuthorizationRule `json:"authorization,omitempty"`
+}
+
+// AccessRuleSource specifies the source of a request.
+//
+// Type must be set to indicate the source type.
+// Similarly, either SPIFFE or Serviceaccount can be set based on the type.
+type AccessRuleSource struct {
```
haiyanmeng, line 107: The subfields of this struct are inconsistent with https://github.com/kubernetes-sigs/kube-agentic-networking/discussions/235#discussioncomment-16519995

api/v1alpha1/accesspolicy_types.go
haiyanmeng: This was discussed on the GH Discussion. I don't see any objection from the GH Discussion.

api/v1alpha1/accesspolicy_types.go
haiyanmeng: SG

api/v1alpha1/accesspolicy_types.go
haiyanmeng: SG

haiyanmeng: SG

api/v1alpha1/accesspolicy_types.go
haiyanmeng: SG

api/v0alpha0/accesspolicy_types.go
ziyue-101: +1, I suggest undo the v0alpah0 changes and re-run the tests.

api/v1alpha1/accesspolicy_types.go
LiorLieberman: you are good! changed!

api/v1alpha1/accesspolicy_types.go
LiorLieberman: I know, decided to have this like we have in Gateway API (where we copied from), in case we ever converge.

api/v1alpha1/accesspolicy_types.go
LiorLieberman: TargetRef validation would be in each control plane, yes. And if invalid status will be posted.

k8s/crds/agentic.networking.x-k8s.io_xaccesspolicies.yaml
```diff
@@ -771,6 +771,822 @@ spec:
...
+                type: array
+                x-kubernetes-list-type: atomic
+            required:
+            - ancestors
+            type: object
+        required:
+        - spec
+        type: object
+    served: true
     storage: true
```
ziyue-101: @LiorLieberman  I think this is the culprit of the tests failures.

Your XAccessPolicy CRD is configured with v1alpha1 as the storage version and v0alpha0 as a served version (but not stored). The conversion strategy between these versions is set to None.

Because the schema changed significantly between v0alpha0 and v1alpha1 (specifically how tools are matched) and there is no conversion webhook, the Kubernetes API server performs a naive conversion when storing the resource as v1alpha1. This causes the tools field to be dropped or lost. 

When the translator encounters an InlineTools rule with an empty tools list, it generates an Envoy RBAC permission that matches anything EXCEPT tools/call.


You either need to

* update the tests to use v1alpha1 api, but the reconciliation part is not there yet.

* implement a conversion webhook to support both v0alpha0 and v1alpah1.

api/v1alpha1/accesspolicy_types.go
david-martin: This is about not mixing targetRefs?
So all can be Gateways, or all can be XBackends, but not a mix of both?

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,302 @@
...
+// AccessPolicyActionType identifies a type of action for access policy.
+// +kubebuilder:validation:Enum=Allow;ExternalAuth
+type AccessPolicyActionType string
+
+const (
+	// ActionTypeAllow is used to identify that the request should be allowed if rules match.
+	ActionTypeAllow AccessPolicyActionType = "Allow"
+
+	// ActionTypeExternalAuth is used to identify that the request should be delegated to an external auth service if rules match.
+	ActionTypeExternalAuth AccessPolicyActionType = "ExternalAuth"
```
david-martin, line 81: Is there a contradiction here?

> the request should be delegated to an external auth service if rules match

and 

> Evaluation logic:
>	// 1. ExternalAuth runs before all other Allow policies.

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,302 @@
...
+	// - From JWTs or other request-level authentication tokens.
+	//
+	// +optional
+	SPIFFE *AuthorizationSourceSPIFFE `json:"spiffe,omitempty"`
+
+	// serviceAccount specifies a Kubernetes Service Account that is
+	// matched by this rule. A request originating from a pod associated with
+	// this Service Account will match the rule.
+	//
+	// The Service Account listed here is expected to exist within the same
```
david-martin, line 136: by trust domain, do we mean same kubernetes cluster, or same namespace?

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,302 @@
...
+	// +optional
+	MCP MCPAttributes `json:"mcp,omitempty"`
+}
+
+// MCPAttributes defines the protocol-specific attributes for MCP authorization.
+type MCPAttributes struct {
+	// Methods is a list of specific MCP functional methods to match.
+	// If specified, only MCP requests with a method
+	// that matches one of these items will be authorized.
+	// If empty or omitted, no method-level whitelisting is applied, meaning all
```
david-martin: let's go with allowlisting here (and at least 1 other place)

https://www.cncf.io/announcements/2021/10/13/inclusive-naming-initiative-announces-new-community-resources-for-a-more-inclusive-future/

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,302 @@
...
+	//
+	PolicyConditionAccepted gwapiv1.PolicyConditionType = "Accepted"
+
+	// This reason is used with the "Accepted" condition when the policy
+	// has been accepted by the controller.
+	PolicyReasonAccepted gwapiv1.PolicyConditionReason = "Accepted"
+
+	// This reason is used with the "Accepted" condition when the policy
+	// was rejected because the maximum number of policies per target was exceeded.
+	PolicyLimitPerTargetExceeded gwapiv1.PolicyConditionReason = "LimitPerTargetExceeded"
```
david-martin, line 254: Do we define a limit or is it implementation specific?

api/v1alpha1/accesspolicy_types.go
david-martin: I was going to ask for some clarification on a user/group scenario with multiple rules having to be allowed, but I think this discussion is moreso about rules across different policies?

Also, we don't have support for groups or spiffe wildcards right now? So that wouldn't be relevant here (which is probably a good motivation for CEL/OIDC)

david-martin (commented):
Some questions so I can better understand the changes from 0 to 1.
In general though, it looks reasonable to me.

/approve

Makefile
```diff
@@ -111,9 +111,12 @@ BOILERPLATE_FILE := hack/boilerplate/boilerplate.generatego.txt
 .PHONY: generate
 generate: manifests deepcopy register clientsets ## Generate manifests, deepcopy code, and clientsets.
 
+# TODO: Remove the python post-processing patch once XAccessPolicy v1alpha1 is fully implemented
+# and flipped to `served: true` instead of v0alpha0.
```
david-martin, line 115: do we have a tracking issue for this?

david-martin:
/hold

pending some q's

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,302 @@
...
+type XAccessPolicy struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// metadata is a standard object metadata.
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// spec defines the desired state of AccessPolicy.
+	// +required
+	Spec AccessPolicySpec `json:"spec"`
```
ryanzhang-oss, line 286: the k8s CRD convention naming would be something like this
```
	Spec XAccessPolicySpec `json:"spec"`
```

I wonder why we have two different naming here?

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,302 @@
...
+	Rules []AccessRule `json:"rules"`
+}
+
+// AccessPolicyActionType identifies a type of action for access policy.
+// +kubebuilder:validation:Enum=Allow;ExternalAuth
+type AccessPolicyActionType string
+
+const (
+	// ActionTypeAllow is used to identify that the request should be allowed if rules match.
+	ActionTypeAllow AccessPolicyActionType = "Allow"
```
ryanzhang-oss, line 78: Shouldn't this be "LocalAuth" when the other Value of Enum is "ExternalAuth"?

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,302 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+// AccessPolicySpec defines the desired state of AccessPolicy.
+//
+// Implementations SHOULD return a regular HTTP formatted response if the policy is enforced against non-MCP traffic.
+// Implementations MAY return a JSON-RPC formatted response if the policy is enforced against MCP traffic.
+// +kubebuilder:validation:XValidation:rule="self.action == 'ExternalAuth' ? has(self.externalAuth) : true",message="externalAuth must be specified when action is set to 'ExternalAuth'"
```
ryanzhang-oss, line 31: I am a bit confused. When the AccessPolicyActionType is ExternalAuth, does the external server do both the authentication and autherization? If so, does it mean that the AccessRule field is not needed in that case? 

Also, how does the external server do Authentication? What identity does the controller use to talk to the external auth sever?

api/v1alpha1/accesspolicy_types.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// MCP defines MCP-specific matching criteria.
+	// If omitted, the policy does not check MCP-level parameters, allowing all MCP traffic that
+	// successfully passes through the matched HTTP routing envelope.
+	// +optional
+	MCP MCPAttributes `json:"mcp,omitempty"`
+}
+
+// MCPAttributes defines the protocol-specific attributes for MCP authorization.
+type MCPAttributes struct {
```
ryanzhang-oss, line 183: Does this imply that the rules are basically based on MCP tool name and parameters?

api/v1alpha1/accesspolicy_types.go
LiorLieberman: Correct. To start simple. I think this came out of a conclusion in extAuth discussion where Backend and Gateway could have been confusing to be together. This may not be required though anymore since we are not releasing this with XBackend support. I still think its smart to include (easier to relax this later)

api/v1alpha1/accesspolicy_types.go
LiorLieberman: >This is not what https://github.com/kubernetes-sigs/kube-agentic-networking/pull/220 proposes. allow policies means inline policies. I am trying to suggest that we define the rules here to address https://github.com/kubernetes-sigs/kube-agentic-networking/issues/237.

We currently propose Allow and ExternalAuth. We dont propose deny in this version given the consensus we got with other folks in the group. This rules are just outlining some guidance on how ExternalAuth and Allow should work **across different policies** (as @david-martin suggested)

David - right - I would really love to see CEL and OIDC support, hopefully you/gui can pick this up soon and we release it as part of 0.2. 

(there is someone thats working on CEL right now, hopefully we incorporate this soon)

api/v1alpha1/accesspolicy_types.go
LiorLieberman: This means that if `action: ExternalAuth` and any rule (defined under the "rules" field) match - it goes to the externalAuth server for evaluation. If `action: Allow` and any rule (defined under the "rules" field) match - the evaluation of this is ALLOW. 

Regardless of that, Evaluation logic between policies on the same target works in a way that if there is ExternalAuth, it before all other Allow policies. If the ExternalAuth server denies it - request is denied. If it allows it, you still need all other policies to allow it (unless they dont exist). 

A concrete example:

action: extAuth
rules:
 - match path /foo
target: Gateway1
---
action: Allow
rules:
 - match header bar
target: Gateway1
 
 If the request is "/foo", ExtAuthServer allows, and header bar is missing -- you still get the request denied.

api/v1alpha1/accesspolicy_types.go
LiorLieberman: its configurable. Often times it would be a cluster identifier, but its not strictly required. 
Here is how it could commonly be: `spiffe://<trust_domain>/ns/<namespace>/sa/<serviceaccount>`

api/v1alpha1/accesspolicy_types.go
LiorLieberman: impl specific

Makefile
LiorLieberman: Planning to fast follow after I add ref implementation support for the new field structure (i.e moving the support to support the v1alpha1 from v0alpha0)

api/v1alpha1/accesspolicy_types.go
LiorLieberman: Good question. Looks like we have not done this in the Gateway community. I have no strong opinion, but its going to be more nice code wise to not call v1alpha1.XAcessPolicySpec. The X prefix serves primarily the purpose of safe upgrades to standard channel and to explicitly indicate that this is an experimental resource 

https://github.com/kubernetes-sigs/gateway-api/blob/main/apisx/v1alpha1/xbackendtrafficpolicy_types.go
https://github.com/kubernetes-sigs/gateway-api/blob/main/apisx/v1alpha1/xmesh_types.go

api/v1alpha1/accesspolicy_types.go
LiorLieberman: The idea is that actions, will evolve to add Deny in the future as well. We already got [some usecases](https://docs.google.com/document/d/11OZY7Sg-xRovqnPedOh7dtLZG2hoztqSaYIY-8tRTq8/edit?tab=t.0) that suggests deny will be needed. 

Then you will have 3 actions. (1) this policy is an ALLOW policy, (2) this policy is a delegate to an external server and (3) this policy is a deny policy (if we ends up adding it in future releases)

api/v1alpha1/accesspolicy_types.go
LiorLieberman: for now yes. The group discussed to add additional support for "Inline" ways to specify other requests attributes like path, headers, etc. 

That said, once CEL support is added you will be able to do more granular things with CEL. You would have AuthorizationRuleTypeCEL (in addition to the AuthorizationRuleTypeInline). See [this slide](https://docs.google.com/presentation/d/1scubdUaDOQzULlyLvwZTK7arttKaGEYsAnFGW5FhgkI/edit?resourcekey=0-pGZ4_x6I6objtgoh2SSn8w&slide=id.g3d85d974727_1_10#slide=id.g3d85d974727_1_10) for the general idea

api/v1alpha1/accesspolicy_types.go
haiyanmeng: https://github.com/kubernetes-sigs/gateway-api/blob/32e498c263a4b450705a21400c9b477aa3e909bb/geps/gep-3779/index.md proposes `AuthorizationPolicy`. We have AccessPolicy in our API. It doesn't makes sense to stay with the GEP proposal.

Makefile
```diff
@@ -111,9 +111,12 @@ BOILERPLATE_FILE := hack/boilerplate/boilerplate.generatego.txt
...
 .PHONY: manifests
 manifests: controller-gen ## Generate CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd paths="./api/..." output:crd:artifacts:config=k8s/crds
+	python3 -c "p='k8s/crds/agentic.networking.x-k8s.io_xaccesspolicies.yaml'; text=open(p).read(); parts=text.split('  - name: v1alpha1'); parts[1]=parts[1].replace('    served: true', '    served: false', 1); open(p,'w').write('  - name: v1alpha1'.join(parts))"
```
bowei, line 119: you should put this in a hack/script.py. (follow up PR)

bowei:
/lgtm

Let's get an initial version in and iterate

PR #261, opened 2026-05-02
Allow counting attached routes even if listener references are unresolved

haiyanmeng:
**What type of PR is this?**

/kind bug

The change ensures that a Gateway listener still counts its attached routes even if its TLS certificate references (Secrets) cannot be resolved. Previously, the code would skip the rest of the processing for that listener immediately upon finding unresolved references, leading to an incorrect count of attached routes.

This gap was identified by the conformance test `TestGatewayAPIConformance/GatewayWithAttachedRoutes/Gateway_listener_should_have_AttachedRoutes_set_even_when_Gateway_has_unresolved_refs`.

```release-note
NONE
```


LiorLieberman (approved):
sweet, thanks @haiyanmeng !
/lgtm

PR #264, opened 2026-05-05
auth: remove CEL vocab standardization

howardjohn:
**What type of PR is this?**

/kind feature

Pulls discussion from https://kubernetes.slack.com/archives/C09P6KS6EQZ/p1776962229367369 into the proposal

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/proposals/0017-DynamicAuth.md
+++ b/docs/proposals/0017-DynamicAuth.md
@@ -370,28 +370,26 @@ spec:
 
 ### Common Expression Language (CEL) for authorization
 
-Common Expression Language (CEL) expressions in AccessPolicy resources have access to a structured context that provides information about the request and the verified identity. Understanding this context is essential for writing effective authorization rules.
+Common Expression Language (CEL) expressions in AccessPolicy resources are evaluated by the implementation that enforces the policy. This proposal standardizes the API field that carries a CEL expression, but does not standardize a portable CEL variable vocabulary at this time.
 
-#### Available Context Variables
+#### CEL Context Vocabulary
 
-##### `request` Object
+Each implementation defines the structured context available to CEL expressions, including available variables, field names, types, missing-field behavior, and protocol-specific objects. Implementations MUST document their supported CEL context for AccessPolicy, and users SHOULD treat CEL expressions as implementation-specific unless a future proposal standardizes a common vocabulary.
 
-The `request` object contains information about the incoming request to the Backend:
+This keeps the API interoperable at the policy shape level while avoiding a premature lowest-common-denominator vocabulary. Existing dataplanes already expose similar request concepts with incompatible semantics. For example, Envoy documents its CEL-facing [attribute vocabulary](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes), while agentgateway documents its [CEL context](https://agentgateway.dev/docs/standalone/latest/reference/cel/cel-context/) and richer [header/query helper behavior](https://github.com/agentgateway/agentgateway/blob/dd018791553486e10e830cc7e6279c965e4051ea/schema/cel-functions.md#header-views):
 
-| Field | Type | Description | Example |
-|-------|------|-------------|---------|
-| `request.method` | `string` | HTTP method | `"POST"` |
-| `request.path` | `string` | Request path | `"/mcp"` |
-| `request.headers` | `map<string, string>` | HTTP headers (lowercase keys) | `{"content-type": "application/json"}` |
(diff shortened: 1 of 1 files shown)


LiorLieberman:
Overall this makes sense to me and feels like a good start. 

/approve

/cc @guicassolato @mikemorris  let me know if you have any particular concerns to start with this.

guicassolato (approved):
I wish implementations could commit to a small set of standardised variables and therefore offer at least that level of interoperability. Whether we like it or not, what goes in a CEL expression becomes part of the API – or a very close extension of it.

By making this part entirely **Implementation-specific**, we're increasing the chances of locking users down to a particular vendor, even when they have only perfectly stated AccessPolicy resources around, or force them to deal with a data migration otherwise unnecessary. It's one more reason to think twice before moving a GatewayClass to a different `controllerName`, I guess.

This change also pushes the project away from the core regarding CEL support[^1]. For references specifically related to auth, please see [^2] and [^3].

[^1]: https://kubernetes.io/docs/reference/using-api/cel/
[^2]: https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#matching-requests-matchconditions
[^3]: https://github.com/luxas/enhancements/blob/conditional_authorization/keps/sig-auth/5681-conditional-authorization/README.md

I believe this change sets a lower bar for interoperability than we might hope for. This is no criticism to @howardjohn's position nonetheless. Given the practical constraints and stakeholder positions, I understand this may be the most viable path forward at this time.

/approve

docs/proposals/0017-DynamicAuth.md
```diff
@@ -370,28 +370,26 @@ spec:
...
-| `request.mcp.params` | `map<string, dyn>` | MCP tool parameters | `{"a": 5, "b": 3}` |
+| Concept | Envoy attribute vocabulary | agentgateway vocabulary | Portability concern |
+|---------|----------------------------|-------------------------|---------------------|
+| HTTP request path without query string | `request.url_path` | `request.path` | The same logical value has different field names. |
+| HTTP request path with query string | `request.path` | `request.pathAndQuery` | The same field name, `request.path`, means different things across dataplanes. |
+| HTTP request headers | `request.headers` is a `map<string, string>` and duplicate values are comma-concatenated | `request.headers` is a header view with direct lookup plus methods such as `join()`, `raw()`, and `split()` | Header cardinality, joining, raw value preservation, case handling, and convenience accessors differ. |
+| MCP request details | Not defined by Envoy's generic HTTP attribute vocabulary | Implementations may expose MCP-specific fields such as method, tool name, and params | Protocol-specific objects are necessarily tied to the implementation and backend type. |
```
david-martin, line 386: > Implementations may expose MCP-specific fields such as method, tool name, and params

Is this intended to be under the agentgateway vocabulary?

david-martin (commented):
I understand the rationale for proposed wording changes.

It does mean we can't write a conformance test that includes a CEL expression.

david-martin:
The [Slack discussion](https://kubernetes.slack.com/archives/C09P6KS6EQZ/p1776962229367369) (as of 12th May) suggests at least one dataplane is willing to adopt standardized variables.
If others follow, we could define a minimal conformance-testable set. Worth waiting for that signal before closing the door on standardization here.

howardjohn:
> The [Slack discussion](https://kubernetes.slack.com/archives/C09P6KS6EQZ/p1776962229367369) (as of 12th May) suggests at least one dataplane is willing to adopt standardized variables.

FWIW they said they would adopt *a vocabulary* but not the one proposed in the current document

mikemorris:
> Worth waiting for that signal before closing the door on standardization here.

I found that Slack thread encouraging for longer-term standardization prospects, but I think the immediate situation is that the specific variables currently have some divergence between dataplanes, and remediating that is neither practical within a control plane, nor a short-term fix for the dataplane implementations, so I do think moving this detail to impl-specific for now makes sense and we should work towards collaborating on longer-term alignment here for the future.

> This change also pushes the project away from the core regarding CEL support

I think the core challenge here is we're not relying on the k8s API server to interpret these CEL expressions, but rather arbitrary dataplane implementations, so consistency with the k8s API server implementation (including the somewhat extensive additional libraries available) can't be expected.

FWIW this feels like basically the same problem as RegEx in Gateway API - we're working on identifying a common subset now in https://github.com/kubernetes-sigs/gateway-api/pull/4764, but didn't block on this for allowing https://gateway-api.sigs.k8s.io/reference/1.5/spec/#pathmatchtype initially.

/lgtm

PR #271, opened 2026-05-12
Align v1alpha1 impl

### LINKED ISSUE #270: v1alpha1 api implementation
add reference implementation support for v1alpha1 api instead of v0alpha0 api.

LiorLieberman:
**What type of PR is this?**
/kind feature

This PR aligns the codebase's XAccessPolicy implementation with the redesigned v1alpha1 specification. High level changes include:
- Move ExtAuth support to top level. (and align with api sematics)
- Support the more native MCP struct instead of "InlineTools"
- Update v0alpha0 references to v1alpha1
- undo unnecessary previous hacks

Fixes #270 

```release-note
upgrade implementation for v1alpha1 XAccessPolicy
```
/cc @bowei


pkg/translator/accesspolicy.go
```diff
@@ -262,26 +273,164 @@ func (t *Translator) translateAccessPolicyToRBAC(accessPolicy *agenticv0alpha0.X
...
+		// Handle ExternalAuth at policy level
+		if accessPolicy.Spec.Action == agenticv1alpha1.ActionTypeExternalAuth && accessPolicy.Spec.ExternalAuth != nil {
+			hash, err := externalAuthUniqueID(accessPolicy.Spec.ExternalAuth)
+			if err != nil {
+				klog.Errorf("Failed to generate unique ID for externalAuth config in AccessPolicy %s/%s: %v", accessPolicy.Namespace, accessPolicy.Name, err)
+				continue
+			}
+			rbacConfig.ShadowRulesStatPrefix = fmt.Sprintf("%s_%s_", constants.ExternalAuthzShadowRulePrefix, hash)
 
-		addPolicyToRBACRules(rbacConfig, policyName, rbacPolicy)
```
david-martin, line 271: @LiorLieberman I think there might be a bug here.
I was trying out these changes locally with the ext auth quickstart example, and it doesn't seem to be working.
I think I've tracked it down to this line being moved inside the else block below.
Before, it was always run for ext auth rules (so shadow & rules would both be added).
Now only shadow rules are added.

Does that make sense? cc @guicassolato in case I'm missing something (this is based on my understanding from https://github.com/kubernetes-sigs/kube-agentic-networking/pull/83#discussion_r2803578343

david-martin:
For anyone else trying this out locally, I hit a couple of issues.
1 a small bug (fixed in #273 )

The other was my own fault.
After running the quickstart, you need to build/reload the controller image so it's running the new code (🤦 )
So after `make quickstart-ollama` (in my case):
```
  make dev-reload-controller
  kubectl delete xaccesspolicy --all -n quickstart-ns
  kubectl apply -f site-src/guides/quickstart/policy/e2e.yaml
```

There may also be an issue with ext auth (comment inline). Will follow up there.

pkg/translator/accesspolicy.go
david-martin: CI e2e failure in TestExternalAuthE2E looks like the same issue, tools getting 403'd before ext_authz runs.

guicassolato: @david-martin's intuition seems correct.

The approach of using the RBAC filter to trigger ext_authz requires not only the shadow rule to be defined (so it sets the stats that govern which ext_authz config to trigger), but also a regular RBAC rule that allows the request to go through, with the same matching conditions. Otherwise, the proxy will deny access on the account of only having regular RBAC rules that do not match the request (or no rule at all).

This is why we had a call to `addPolicyToRBACShadowRules` only when `ExternalAuth` and a call to `addPolicyToRBACRules` regardless. The regular rule allows the request, while the shadow rule delegates the decision to ext_authz.

guicassolato (changes requested):
I'm trying the PR with the ext auth quick start now. All tool/call requests are returning `Access to this tool is forbidden (403)`. 

In the proxy logs, I see `[debug][rbac] enforced denied, matched policy none`.

In the proxy config, all MCP `tools/call` permissions are wrapped within a `not_rule`, which I didn't expect. I also didn't see any shadow rules or ext_authz filter configs at all. Yet, an `ExternalAuth` XAccessPolicy and an `Allow` another exist, targeting two different backends, and they are both `Accepted`.

I'm looking into it.

guicassolato:
> I'm trying the PR with the ext auth quick start now. All tool/call requests are returning `Access to this tool is forbidden (403)`.
> 
> In the proxy logs, I see `[debug][rbac] enforced denied, matched policy none`.
> 
> In the proxy config, all MCP `tools/call` permissions are wrapped within a `not_rule`, which I didn't expect. I also didn't see any shadow rules or ext_authz filter configs at all. Yet, an `ExternalAuth` XAccessPolicy and an `Allow` another exist, targeting two different backends, and they are both `Accepted`.
> 
> I'm looking into it.

My bad. I was running the wrong image :/

I think caught another bug, but not this one. Sorry for the noise.

pkg/translator/accesspolicy.go
```diff
@@ -262,26 +402,152 @@ func (t *Translator) translateAccessPolicyToRBAC(accessPolicy *agenticv0alpha0.X
...
+			addPolicyToRBACRules(rbacConfig, policyName, allowAllPolicy)
+		} else {
+			// Action is Allow
+			// Ensure permissions is never empty to satisfy Envoy's schema validation (min_items: 1).
+			// Spec guidance: If omitted, all access from the specified source is allowed.
+			if len(rbacPolicy.GetPermissions()) == 0 {
+				rbacPolicy.Permissions = []*rbacconfigv3.Permission{buildAnyPermission()}
+			}
+			addPolicyToRBACRules(rbacConfig, policyName, rbacPolicy)
+		}
```
guicassolato, line 374: We should set an allow policy with the **same matchers as the external auth** shadow rules one. Otherwise, we'll be allowing all principals and all requests, including the ones that will not trigger external auth.

```suggestion
		} else {
			// Action is Allow
			// Ensure permissions is never empty to satisfy Envoy's schema validation (min_items: 1).
			// Spec guidance: If omitted, all access from the specified source is allowed.
			if len(rbacPolicy.GetPermissions()) == 0 {
				rbacPolicy.Permissions = []*rbacconfigv3.Permission{buildAnyPermission()}
			}
		}
		addPolicyToRBACRules(rbacConfig, policyName, rbacPolicy)
```

pkg/translator/accesspolicy.go
LiorLieberman: Thanks for pointing out. Resolving this comment and will respond on the more recent ones on this matter

LiorLieberman:
Ok, thanks all for the review and great findings! I am providing a summary of the approach this PR takes after a bunch of research and discussions; 

* ShadowRules exist in the repo for the purpose of conditionally trigger external callouts per route. The way it works is that the shadowRules are evaluated without actual enforcement and they set a stat, and the ext_auth filter is conditionally triggered based on this stat. 

* For ExternalAuth policies, the rules define request matchers that are subjected to external auth. If a request doesn't match, we want it to not execute the ext_auth filter and proceed to regular Allow policies. This is why so that non-matching requests **wont** get implicitly denied by it.

* We have one RBAC Filter per {target, policy action} tuples

* ExternalAuth policy is always executed before Allow policies. Filter chain order is as follows: `MCP → Gateway ExtAuth (shadowRules) → Backend ExtAuth (shadowRules) → Ext_auth→ Gateway Allow RBAC → Backend Allow RBAC → Router`

If no ExternalAuth policy is specified, no ShadowRules will be generated.

pkg/translator/accesspolicy.go
LiorLieberman: Thanks Gui, following up our offline chat closing this and commented below

guicassolato (approved):
As discussed offline, there are still quite a few things we can improve from here, such as (tho no limited to):
1. No longer creating the default RBAC policies associated to the ext auth config — they are a no-op since the removal of the condition to trigger ext auth on `tools/call` requests only. 
2. Some duplication of code we can easily remove in a refactoring later on
3. Clarify the behaviour in the case of ExternalAuth policies targeting simultaneously the Gateway and a Backend
4. Minor fixes to the docs and quickstart guides to reflect the new behaviour.

Nevertheless, I’m happy to unblock the PR whose work is needed and in line with several points agreed with the community on how to enhance the XAccessPolicy API.

Thanks, @LiorLieberman, for the hard work!

PR #278, opened 2026-05-18
Add conformance reports

LiorLieberman:
/kind test

add conformance report structure

```release-note
NONE
```


version/version.go
```diff
@@ -0,0 +1,34 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package version
+
+var (
+	// CommitSHA is the git hash of the latest commit in the build.
```
bowei, line 20: you might want to document that these are set by the build?

bowei:
Commit messages have some random stuff in them (tag and what looks like a session identifier)

pkg/generator/main.go
```diff
@@ -0,0 +1,138 @@
+/*
```
bowei, line 1: Add top-level comment on this file why we need this generator.

bowei:
Changes lgtm, but you might want to document some of the things such as the need for a special rewrite of the datetime. Do other projects need this or is this somehow special to what we are doing here?

david-martin:
@bowei I didn't intend for my lgtm to merge this (thought an approve was also needed).
Your comments are still valid, and something we can follow up on after.

Issue #277, opened 2026-05-15
makes rules field optional


LiorLieberman:
**What would you like to be added**:
makes rules field optional

**Why this is needed**:
mainly to enable global extAuth callouts and have more flexibility in ALLOW policies as well - e.g empty rules == deny all (deny-by-default).

david-martin:
Closing as done in #279

PR #279, opened 2026-05-18
implement optional rules and configurable base MCP method matching

### LINKED ISSUE #276: Add a convenient way to match on non-access MCP protocol methods
a field to indicate that an XAccessPolicy author wants to match on basic non-access specific MCP methods. Namely:
* initialize
* completion/
* logging/
* notifications/
* ping

**Why this is needed**:

Without it, anyone that deploys a policy, even if they allow specific tools and methods, their mcp traffic is likely to break without allowing `initialize` for example

LiorLieberman:
/kind feature

This PR makes the `rules` field in AccessPolicy optional and introduces a new enum field MCPBaseProtocolMethodsOption to control the injection of default allowed rules for basic MCP operations.

TODO: what about HTTP GET (Required for SSE Transport), HTTP DELETE (Required for Session Close), and tools/list? should we allow those as part of the base methods?

Fixes #276, #277

```release-note
rules are now optional and new mcpBaseProtocolMethodsOption is implemented
```

### DIFF
--- a/api/v1alpha1/accesspolicy_types.go
+++ b/api/v1alpha1/accesspolicy_types.go
@@ -87,10 +87,10 @@ const (
 type MCPBaseProtocolMethodsOption string
 
 const (
-	// SKIP_BASE_PROTOCOL_METHODS skips matching on the base MCP protocol methods.
-	SKIP_BASE_PROTOCOL_METHODS MCPBaseProtocolMethodsOption = "SKIP_BASE_PROTOCOL_METHODS"
-	// MATCH_BASE_PROTOCOL_METHODS matches on the base MCP protocol methods.
-	MATCH_BASE_PROTOCOL_METHODS MCPBaseProtocolMethodsOption = "MATCH_BASE_PROTOCOL_METHODS"
+	// MCPBaseProtocolMethodsOptionSkip skips matching on the base MCP protocol methods.
+	MCPBaseProtocolMethodsOptionSkip MCPBaseProtocolMethodsOption = "SKIP_BASE_PROTOCOL_METHODS"
+	// MCPBaseProtocolMethodsOptionMatch matches on the base MCP protocol methods.
+	MCPBaseProtocolMethodsOptionMatch MCPBaseProtocolMethodsOption = "MATCH_BASE_PROTOCOL_METHODS"
 )
 
 // AccessRule specifies an authorization rule for a specified target.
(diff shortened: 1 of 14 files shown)


david-martin:
Would it make sense to split optional rules into its own PR?
The optional rules change is well-bounded and straightforward, relaxing the API validation, handling empty rules in the translator, updating the docs. I think that can land quickly without much debate.

The MCPBaseProtocolMethodsOption side has some open design questions that I think are worth discussing a bit more before we commit to an API shape.

(For the optional-rules PR, I'd suggest keeping the existing common policy auto-injection intact (HTTP GET, session close, initialize/list-tools) so there's no behavioral change beyond making rules optional)

On the base protocol methods design, building on my comment on #276, I think there's a distinction between capability discovery and protocol:

- Protocol plumbing (initialize, initialized, ping, notifications/*) — these are part of the MCP handshake/lifecycle. Is there any realistic use case for blocking them per-identity, and requiring policy authors to opt them in?.
- Capability discovery (tools/list, and eventually prompts/list, resources/list), these feel more access-specific. You could want an identity that can discover and use prompts but shouldn't even see what tools are available. Bundling tools/list into "base protocol" would close off that option.
- HTTP transport (GET for SSE, DELETE for session close), transport-level, should probably stay implicitly allowed.

What do you think? Happy to discuss further here or on #276.

david-martin:
@LiorLieberman Thinking more on the enum approach after our call, here's my take on what `base` should include:

**Lifecycle methods per the [MCP spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle):**
- `initialize`
- `notifications/initialized`

**Utilities:**
- `ping`
- `notifications/progress` (typically server→client, but worth covering)

**I'd also argue `tools/list` belongs in `base`.** A client can't discover what tools are available without it — the only alternative is out-of-band discovery, which is a [known gap in the protocol](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/908). Given that tool authorization is the core value prop of AccessPolicy, requiring users to explicitly allow `tools/list` feels unnesessary. Including it in `base` means a policy with just a `tools/call` rule works out of the box.

In future we could define additional enums that align with protocol capability groups — e.g. `completion`, `logging`, `roots`, `sampling`, `elicitation`, `notifications` — but `base` + `tools/list` is the right scope for 0.1.

LiorLieberman:
discussed at the community meeting - 

`completion` and `logging` are _likely_ useful to be there as well, as some clients are likely implemented with those. 

for notifications - I'd argue that we should have them all in that "Base" enum - as I assume some clients also just implement it and would maybe not working well without it - https://modelcontextprotocol.io/docs/learn/architecture#client-response-to-notifications

@david-martin thoughts?

david-martin:
> I'd argue that we should have them all in that "Base" enum

sounds reasonable
So `notifcations/*`?

I'm good with completion and logging as well.

AccessPolicy is primarily about restricting tool calling, so enabling expected/base functionality make sense.
admins can always chose to define specific methods and drop the enum if they're not happy with it.

LiorLieberman:
Adding tools/list then

bexxmodd:
iirc there's agreement in the meeting too, so going to lgtm as it is needed for the conformance PR that comes next.

/lgtm

david-martin:
/lgtm

I've a couple of follow up q's (following up on slack), but this is good based on discussions on weekly call and async.

PR #280, opened 2026-05-20
Implement gateway-level authorization tracing for agent tool calls

evaline-ju:
**What type of PR is this?**

/kind feature

Adds OpenTelemetry tracing to the gateway that captures authorization decisions as span attributes, enabling platform teams to audit agent tool call authorization from traces. Tempo and Grafana happened to be chosen here but any tracing backends able to process OpenTelemetry traces can be substituted.

Controller updates
- Configure OTel tracer in Envoy bootstrap with collector cluster endpoint
- Add shadow RBAC rules alongside enforcement rules to emit security_rule.name on traces
- Add Lua HTTP filter that derives security event attributes from RBAC and MCP dynamic metadata at the gateway level
- Add custom tags for security_rule.name and peer.spiffe.id from RBAC metadata and Lua filter's namespace
- Emit `security_rule.implicit: true` on default-deny spans (in accordance with #312 updates), replacing the previous `security_rule.match` attribute

ADK agent updates
- OTel instrumentation
- Add callbacks for gen_ai.tool.call spans that link agent traces to gateway spans via traceparent

Observability stack
- Helm chart packaging OTel collector, Tempo for trace storage, Grafana pre-configured with Tempo datasource
- Notes: attribute derivation happens with the Lua filter not in Otel collector transforms now, this is not deployed with an operator at this moment and would require future integration

Note: The gateway request span is named `mcp.gateway.request` (matching the proposal) but authorization decision attributes (security_rule.name, event.action, security_rule.implicit) are on the parent span rather than in separate mcp.authorization / mcp.authorization.rule child spans, due to Envoy's tracing model not supporting custom child span creation from filters. This could potentially be split in the collector but would couple the logic of the controller and observability stack - left as future work.

Test steps
```
 Prerequisites: Kind cluster kan-quickstart running with the quickstart deployed (tested with ollama)

  1. Build and patch controller
  docker build -t agentic-networking-controller:trace-test .
  kind load docker-image agentic-networking-controller:trace-test --name kan-quickstart
  kubectl set image deployment/agentic-net-controller -n agentic-net-system
  manager=agentic-networking-controller:trace-test
  kubectl rollout status deployment/agentic-net-controller -n agentic-net-system

  2. Deploy observability stack
  helm install observability-example charts/observability-example \
    --namespace quickstart-ns
  kubectl wait --for=condition=available --timeout=120s \
    deployment/otel-collector deployment/tempo deployment/grafana -n quickstart-ns

  3. Restart gateway to pick up new xDS config (with Lua filter)
   GATEWAY_DEPLOY=$(kubectl get deployment -n quickstart-ns \
    -l "gateway.networking.k8s.io/gateway-name=agentic-net-gateway" \
    -o jsonpath='{.items[0].metadata.name}')
  kubectl rollout restart deployment/${GATEWAY_DEPLOY} -n quickstart-ns
  kubectl rollout status deployment/${GATEWAY_DEPLOY} -n quickstart-ns

  4. Build and patch agent
  docker build -t quickstart-adk-agent:trace-test site-src/guides/quickstart/adk-agent/
  kind load docker-image quickstart-adk-agent:trace-test --name kan-quickstart
  kubectl set image deployment/adk-agent -n quickstart-ns adk-agent=quickstart-adk-agent:trace-test

  # Option A: Apply the full deployment (includes all env vars and fieldRef metadata)
  # Requires the hf-secret Secret to exist (referenced by the HF_TOKEN env var)
  kubectl apply -f site-src/guides/quickstart/adk-agent/deployment.yaml

  # Option B: Patch the existing deployment (sufficient for tracing, skips fieldRef vars)
  kubectl set image deployment/adk-agent -n quickstart-ns adk-agent=adk-agent:latest
  kubectl set env deployment/adk-agent -n quickstart-ns \
    OTEL_EXPORTER_OTLP_ENDPOINT="http://otel-collector.quickstart-ns.svc.cluster.local:4318" \
    OTEL_SERVICE_NAME="adk-agent" \
    ADK_ENABLE_OTEL="true" \
    OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED="true" \
    OTEL_PROPAGATORS="tracecontext,baggage"

  kubectl rollout status deployment/adk-agent -n quickstart-ns

  5. Verify traces
  kubectl port-forward -n quickstart-ns svc/grafana 3000:3000 &
  kubectl port-forward -n quickstart-ns svc/adk-agent-svc 8081:80 &
  - Trigger a tool call via http://localhost:8081/dev-ui/?app=mcp_agent
  - Open Grafana at http://localhost:3000 (admin/admin) → Explore → Tempo
  - Search by tag: security_rule.name=tools-for-adk-agent-sa or event.action=allow
```

Assisted by Claude Code with Opus 4.6

Begins to implement observability proposal in #33 

```release-note
Adds observability stack integration to the reference implementation
```


david-martin:
@evaline-ju reading through https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/rbac/v3/rbac.proto, and doing some digging to better understand the problem, I can see it says "A match occurs when at least one policy matches the request". And like you say, the code breaks on the 'first match' https://github.com/envoyproxy/envoy/blob/c784df87dd7c5806ff66785fe1af08041b431d00/source/extensions/filters/common/rbac/engine_impl.cc#L108

Either way, shadow rules or a separate filter or wasm (as suggested by Daniel in the meeting) seem like the viable options to get the results of every rule for auditing/observability.



Question to other (non envoy) gateway maintainers:

> Envoy's RBAC filter breaks on first policy match and only reports one effective_policy_id, so if you have multiple authorization rules per request, you can't observe which ones passed/failed individually. Does your gateway have the same limitation, or can you get per-rule evaluation results?

@kflynn @howardjohn I'd be interested in your thoughts on this ^^

@LiorLieberman anyone else you can think of tagging here?

howardjohn:
I don't quite understand. if two rules allow are you trying to emit both of them? one of them (random? some determining factor on which is "first ")? Your comment seems to suggest all of them but the PR and proposal seem to say 1?

evaline-ju:
@howardjohn What exists for the current agentic networking reference implementation + this PR: there’s the one RBAC filter with multiple policies (one per `XAccessPolicy` rule), and it seems Envoy evaluates these and can only report the single winning rule match - so I can only output a span for the rule that determined the overall outcome (allow/deny), not for the (other) rules that were evaluated but didn't match.

What was envisioned in [proposal 33](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/main/docs/proposals/0033-Observability.md) at the time was we could emit a span for every single evaluated rule (whether or not it matched) for observability and auditing. This isn’t for whether multiple rules ‘allow’ but more on which rules were even considered (including those that did not match) and which one made the final decision (such as the allow). 

We're mostly curious if this is an Envoy limitation vs. other gateways

howardjohn:
I think its a fundamentally unscalable design regardless of the implementation. We definitely have users with rules in the hundreds.. we prob don't want to  evaluate and emit traces for all of those.

pkg/translator/accesspolicy.go
```diff
@@ -371,6 +372,7 @@ func (t *Translator) translateAccessPolicyToRBAC(accessPolicy *agenticv1alpha1.X
 				rbacPolicy.Permissions = []*rbacconfigv3.Permission{buildAnyPermission()}
 			}
 			addPolicyToRBACRules(rbacConfig, policyName, rbacPolicy)
+			addPolicyToRBACShadowRules(rbacConfig, policyName, rbacPolicy)
```
LiorLieberman, line 394: whats the rationale for this?

pkg/translator/accesspolicy.go
```diff
@@ -511,10 +513,22 @@ func addPolicyToRBACRules(rbacConfig *rbacv3.RBAC, policyName string, policy *rb
...
+// Uses RBAC_ALLOW so shadow results mirror enforcement: rule match → allowed, no match → denied.
 func addPolicyToRBACShadowRules(rbacConfig *rbacv3.RBAC, policyName string, policy *rbacconfigv3.Policy) {
 	if rbacConfig.GetShadowRules() == nil {
 		rbacConfig.ShadowRules = &rbacconfigv3.RBAC{
-			Action:   rbacconfigv3.RBAC_DENY, // the action for the shadow rule doesn't really matter in this case since we only use it to trigger ext_authz from emitted stats
+			Action:   rbacconfigv3.RBAC_ALLOW,
+			Policies: map[string]*rbacconfigv3.Policy{},
+		}
+	}
+	rbacConfig.ShadowRules.Policies[policyName] = policy
```
LiorLieberman, line 638: I probably need a better comment here to understand better the logic. It looks like we are now leveraging the shadow rules not only for extAuthz but also for the tracing? Can you clarify and potentially have a comment that explains this for future readers/reviewers?

pkg/translator/accesspolicy.go
LiorLieberman: I think I got it - we now have extAuth as one shadowrules and we also leverage shadow rules for tracing (mirroring non-shadow rules) separately?

pkg/translator/listener.go
```diff
@@ -433,6 +438,110 @@ func buildUDPFilterChain(lis gatewayv1.Listener) (*listener.FilterChain, error)
 	}, nil
 }
 
+// buildTracingConfig creates tracing configuration with custom tags that expose
+// shadow RBAC metadata, caller identity, and derived security event attributes on gateway spans.
+func buildTracingConfig() *hcm.HttpConnectionManager_Tracing {
```
LiorLieberman: @gkhom would appreciate your review here

pkg/translator/httproute.go
```diff
@@ -89,6 +89,11 @@ func (t *Translator) translateHTTPRouteToEnvoyRoutes(
 				Match:                  routeMatch,
 				RequestHeadersToAdd:    headersToAdd,
 				RequestHeadersToRemove: headersToRemove,
+				// Sets the span name for gateway traces; also used by the OTel
+				// collector transform to identify gateway spans by name.
+				Decorator: &routev3.Decorator{
+					Operation: "ingress",
```
LiorLieberman: why ingress?

LiorLieberman (commented):
Thanks for all the work @evaline-ju, left a few initial comments. 

/cc @gkhom for additional more specific review. 

Two main questions/comments;

1. Following our meeting today and on the challenge described in the comments and the meeting notes - I think that we dont necessarily need to indicate all possible policies/rules that were considered during a decision, but rather the decision itself and the rule that "caused it". (first "winning" rule match that allows it is fine in my opinion). For the case where nothing allowed the request and it went to the "default-deny" rule, IMO its ok to only audit that rule. If someone needs to do an more-in-depth assessment what policies exist and why something was not allowed then can always use specific tooling available. 

2. What is the exact reason we choose to leverage lua here? 

also /cc @kyessenov to provide chime in with more in-depth envoy expertise

pkg/translator/httproute.go
evaline-ju: This just sets the envoy route decorator to become the span name for gateway traces, where the gateway handled inbound traffic to the backends, but if "ingress" is confusing, we could rename to something else (like `gateway` etc.?)

pkg/translator/accesspolicy.go
evaline-ju: Yup 👍 the `addPolicyToRBACShadowRulesForExtAuth` for ext_authz uses rbac_deny so a matching request writes out the dynamic metadata to be used as a gate on the external auth, while this one uses rbac_allow to duplicate the enforcement/allow rules so shadow stats expose which rule matched for tracing

pkg/translator/accesspolicy.go
evaline-ju: based on the previous comment as well I've updated a few comments to hopefully clarify this

evaline-ju:
@LiorLieberman 
> * Following our meeting today and on the challenge described in the comments and the meeting notes - I think that we dont necessarily need to indicate all possible policies/rules that were considered during a decision, but rather the decision itself and the rule that "caused it". (first "winning" rule match that allows it is fine in my opinion). For the case where nothing allowed the request and it went to the "default-deny" rule, IMO its ok to only audit that rule. If someone needs to do an more-in-depth assessment what policies exist and why something was not allowed then can always use specific tooling available.

Based on the discussion I’m fine with rolling with the “which rule made the final decision?” implementation as is currently. I assume we may want to update or caveat the original proposal?

> * What is the exact reason we choose to leverage lua here?

I was mainly looking for a way to move some of the logic from OTel collector transform config (now more optional with the separately installed observability stack) to the controller itself. This logic was doing things like setting the authorization trace attributes (e.g. `event.*` ones) by extracting and propagating attributes from the RBAC shadow metadata. @david-martin had suggested a Lua filter as an Envoy way to help do this, and I thought it would be less overhead than a separate wasm binary but open to other options

pkg/translator/listener.go
```diff
@@ -433,6 +438,110 @@ func buildUDPFilterChain(lis gatewayv1.Listener) (*listener.FilterChain, error)
...
+			buildMetadataTag("event.outcome", constants.ObservabilityMetadataNamespace, "event.outcome"),
+			buildMetadataTag("event.kind", constants.ObservabilityMetadataNamespace, "event.kind"),
+			buildMetadataTag("event.category", constants.ObservabilityMetadataNamespace, "event.category"),
+			buildMetadataTag("mcp.method.name", constants.ObservabilityMetadataNamespace, "mcp.method.name"),
+			buildMetadataTag("component", constants.ObservabilityMetadataNamespace, "component"),
+		},
+	}
+}
+
+const observabilityLuaScript = `
```
gkhom, line 463: I understand the need for the Lua filter, but I'm wondering if it would be better to avoid Lua in the proxy and handle this logic in the OTel Collector instead?

Rather than hardcoding the `tools/call` check and `if/else` logic in Envoy, we could just export the raw RBAC/MCP metadata as trace tags. Then the OTel Collector's `transform` processor (OTTL) can derive the `event.action` and `event.outcome` attributes.

This would keep the proxy lightweight and avoid coupling the data plane to specific MCP method names. It would also centralize the telemetry transformation in the collector pipeline. What do you think?

evaline-ju: The logic in the OTel collector transform as suggested had been present in an earlier iteration here, but I had been moved it out to the Lua filter to make the OTel collector itself fully optional, as part of the completely optional and separately installed observability stack for the reference implementation. The thought was if some of the transformation logic lived in the collector’s transform, then the collector would naturally become a hard dependency for those trace attributes. 

The compromise could be to as mentioned to just export the raw RBAC shadow metadata, and we document the collector config necessary to derive the attributes without necessitating the stack inclusion, at the tradeoff of the proposal-documented attributes not being present without the collector.

gkhom: That's a fair point. If the aim is to keep the collector as an optional component then using a Lua filter seems reasonable. An alternative could be wasm, but that would require a bunch of additional machinery that is likely not worth it.

pkg/translator/listener.go
```diff
@@ -433,6 +438,110 @@ func buildUDPFilterChain(lis gatewayv1.Listener) (*listener.FilterChain, error)
 	}, nil
 }
 
+// buildTracingConfig creates tracing configuration with custom tags that expose
+// shadow RBAC metadata, caller identity, and derived security event attributes on gateway spans.
+func buildTracingConfig() *hcm.HttpConnectionManager_Tracing {
+	return &hcm.HttpConnectionManager_Tracing{
+		RandomSampling:    &typev3.Percent{Value: 100.0},
+		ClientSampling:    &typev3.Percent{Value: 100.0},
+		OverallSampling:   &typev3.Percent{Value: 100.0},
```
gkhom: Is there a possibility to make the sampling rate(s) configurable? Hardcoding it to capture 100% of traces is generally considered as an anti-pattern in production.

pkg/translator/listener.go
```diff
@@ -433,6 +438,110 @@ func buildUDPFilterChain(lis gatewayv1.Listener) (*listener.FilterChain, error)
...
+  end
+
+  local match = (rule_name ~= "")
+  meta:set("agentic_obs", "security_rule.match", match)
+  meta:set("agentic_obs", "component", "envoy.gateway")
+
+  if match then
+    meta:set("agentic_obs", "event.action", "allow")
+    meta:set("agentic_obs", "event.outcome", "success")
+  elseif mcp_method == "tools/call" then
```
gkhom: Just noticed that we're only setting the `deny` action and event categories (`event.kind` and `event.category`) specifically for the `tools/call` method here. 

Does this mean auth denials for other MCP methods will be missing these attributes? It might be worth generalizing this to cover all MCP methods so we don't silently drop auth telemetry for them.

pkg/translator/listener.go
```diff
@@ -433,6 +438,110 @@ func buildUDPFilterChain(lis gatewayv1.Listener) (*listener.FilterChain, error)
...
+    meta:set("agentic_obs", "event.action", "deny")
+    meta:set("agentic_obs", "event.outcome", "success")
```
gkhom: Looks like we currently always set `event.outcome` to `success`. How do we intend to flag policy execution failures (especially when policies are evaluated remotely)? Or is this out-of-scope for this PR?

pkg/translator/listener.go
gkhom: Added my 2 cents.

pkg/translator/listener.go
evaline-ju: Thanks for pointing this out. I was focused only on the `tools/call` auth but can see how this should apply for the other MCP methods. Made an update here

pkg/translator/listener.go
evaline-ju: I agree this would be anti-pattern for production and am aware the reference implementation here has been shared with many disclaimers that it’s not production-ready. 

The problem was with this reference implementation there didn’t seem to be a straightforward place to pipe configurable values through - I think it’d likely have to be a gateway annotation (which seems too informal) or new CR field? Happy to hear any suggestions here but currently I’d prefer to track this separately in a follow-up especially if it’d affect other APIs.

pkg/translator/listener.go
evaline-ju: yeah I’ve kept the policy execution failures out of scope since with ext authz failures we don’t get the dynamic metadata for the Lua filter to inspect. I’ve documented this limitation in a code comment for now. We would need a follow-up in the form of a separate filter or plugin likely for the full filter chain state

Rul1an:
The deciding-rule direction reads cleanly to me. With rules running into the hundreds on one target, a span per evaluated rule on the request path would not pay for itself, so the scalability concern raised earlier holds. Recording the rule that caused the outcome, and the default-deny rule when nothing allowed, answers the audit question most people actually have.

What I would make explicit in the span semantics is what that recorded rule is and is not. A span carrying the deciding rule is decisive attribution: it says which rule produced this outcome. It is not a record that every rule was evaluated and agreed, because first-match-wins means the others never ran to a verdict. Those are two different questions, and someone reading `security_rule.name` weeks later can easily take the first for the second.

Keeping decision attribution separate from evaluation coverage costs almost nothing here. A small marker on the span, saying the attribution is first-match and that the absence of sibling rule spans is by design rather than a claim that no other rule would have matched, lets an auditor read the trace for exactly what it is. The deeper "which policies exist and why was this not allowed" assessment then stays the separate, opt-in thing it should be, and the default path does not quietly imply a completeness it cannot show.

Happy to help word the attribute or a short note in proposal 0033 if that is useful.

pkg/translator/listener.go
evaline-ju: follow-up: will look into seeing if we can capture this as configurable arg https://github.com/kubernetes-sigs/kube-agentic-networking/blob/main/k8s/deploy/deployment.yaml#L120

evaline-ju: random and overall sampling is now configurable through controller args

LiorLieberman:
Hi Evaline, if this is ready for review - please dont forget to mark as ready

evaline-ju:
@LiorLieberman thanks for reminding! I have put some related proposal updates up in https://github.com/kubernetes-sigs/kube-agentic-networking/pull/312 so I wanted to get that one in before making any more other impl updates here

LiorLieberman:
hey 👋 , 
Not that #312 is merged, anything else thats blocking here?

LiorLieberman (commented):
I think there has been enough discussion mostly agreement here and in the community meetings.

Thanks @evaline-ju !
/lgtm
/approve

Issue #272, opened 2026-05-13
Quickstart fails: deploy_controller function collision between lib.sh and run-quickstart.sh


david-martin:
Running any quickstart variant (`make quickstart`, `make quickstart-ollama`, `make quickstart-gemini`) fails with:

```
dev/ci/lib.sh: line 215: $1: unbound variable
```

`install_metallb_step` in `run-quickstart.sh` sources `dev/ci/lib.sh`, which defines `deploy_controller(tag, namespace)`. This overwrites the quickstart's own `deploy_controller()` (no args). When the quickstart later calls `deploy_controller`, it hits the lib.sh version and fails because no arguments are passed.

Introduced by the combination of:
- PR #251 (April 23): added `source dev/ci/lib.sh` to the quickstart
- PR #263 (May 6): added `deploy_controller` to `lib.sh`

Fix: run the source in a subshell so lib.sh functions don't leak into the quickstart's scope.

PR #284, opened 2026-05-22
Add quickstart smoke test to e2e suite

### LINKED ISSUE #281: Add quickstart smoke test to e2e suite
## What would you like to be added

An e2e test that validates the quickstart flow works end-to-end: applying the quickstart manifests (Gateway, HTTPRoutes, XBackends, XAccessPolicies), deploying the MCP server, and verifying that tool calls are allowed/denied according to the policies.

This would catch regressions like the function collision bug (#272) which both broke the quickstart but weren't caught by existing e2e tests.

## What should be tested

Using the quickstart policy manifests from `site-src/guides/quickstart/policy/e2e.yaml`:

- `initialize` - **allowed** (base protocol method)
- `tools/list` - **allowed** (base protocol method)
- `get-sum` on local MCP backend - **allowed** (in inline allowlist)
- `get-tiny-image` on local MCP backend - **allowed** (in inline allowlist)
- `echo` on local MCP backend - **denied** (not in allowlist)

Optionally, with external auth quickstart manifests:
- Tool call to remote MCP backend with allowed repo - **allowed**
- Tool call to remote MCP backend with disallowed repo - **denied**

## Options for running this

**Option A: Add to existing e2e suite** (`tests/e2e/controller_test.go`)

Add a new `TestQuickstartE2E` function that applies the quickstart manifests directly. The existing `run-e2e.sh` runner already calls `setup_cluster_with_controller`, so the cluster and controller are ready. The test would apply `site-src/guides/quickstart/policy/e2e.yaml` and the MCP server deployment, then run assertions using the same `callTool`/`runKubectl` helpers.

Pros: fits into existing CI (`pull-kube-agentic-networking-e2e`), reuses test infrastructure.
Cons: quickstart manifests hardcode `quickstart-ns` namespace which differs from the e2e namespace (`e2e-test-ns`).

**Option A variant: Symlink or copy quickstart manifests into testdata**

Instead of referencing `site-src/` paths, copy the quickstart manifests into `tests/e2e/testdata/` (or symlink). Avoids cross-directory dependencies but risks drift between the test copy and the real quickstart manifests.

**Option B: New test suite that runs the quickstart script**

A separate test or script (`dev/ci/run-quickstart-smoke.sh`) that actually runs `run-quickstart.sh` (without the agent/port-forward steps), then runs targeted tool call assertions against the gateway.

Pros: tests the actual quickstart script, catches script-level bugs (like #272).
Cons: needs a new Prow job, slower, harder to run assertions without the Go test framework.

**Option C: Hybrid - Go test using quickstart manifests**

A Go test in `tests/e2e/` that applies the quickstart YAML files directly (not via the script) but uses the quickstart namespace and manifest paths. This validates the manifests are correct without testing the bash script itself.

Pros: catches manifest/policy regressions, runs in existing CI.
Cons: doesn't catch script bugs.

Options A or C seem most practical - they fit into existing CI and catch the most important regressions (policy/manifest correctness). Script-level bugs like #272 are rarer and can be caught by manual testing.

HarshithaMS005:
**What type of PR is this?**
/kind test

The quickstart guide (site-src/guides/quickstart/) is the main way users try Agentic Networking, but nothing in CI exercised those manifests end-to-end. Existing e2e tests use custom testdata/ resources and a single /mcp route, so regressions in the quickstart YAML or policy wiring (for example [#272](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/272)) could break the quickstart without failing CI.
This PR adds TestQuickstartE2E to the existing e2e suite. It applies the real quickstart MCP server deployment and policy/e2e.yaml (Gateway, HTTPRoutes, XBackends, XAccessPolicies), then checks MCP calls through /local/mcp and /remote/mcp match the documented allow/deny behavior. It runs in the same Kind-based e2e job (pull-kube-agentic-networking-e2e) as the other controller tests, so manifest and policy mistakes are caught before users hit them in the quickstart.

Fixes #281 

```release-note
NONE
```

### DIFF
--- a/dev/ci/run-e2e.sh
+++ b/dev/ci/run-e2e.sh
@@ -32,6 +32,8 @@ main() {
 
   header "Running E2E tests"
   # Requirements: K8s v1.35+, PodCertificateRequest/ClusterTrustBundle enabled, and KAN Controller running with --enable-agentic-identity-signer=true.
+  # Remote quickstart MCP tests call mcp.deepwiki.com and are skipped in CI by default.
+  export SKIP_REMOTE_MCP="${SKIP_REMOTE_MCP:-true}"
   cd tests && go clean -testcache && go test -v ./e2e/...
 }
 
--- a/tests/e2e/controller_test.go
+++ b/tests/e2e/controller_test.go
@@ -50,6 +50,8 @@ const (
 	testClientCertPath = "/tmp/mtls-client/tls.crt"
 	testClientKeyPath  = "/tmp/mtls-client/tls.key"
 	testClientCAPath   = "/tmp/mtls-client/ca.crt"
+
+	defaultMCPPath = "/mcp"
 )
 
 // TestControllerE2E verifies the core functionality of the agentic networking controller including:
@@ -739,13 +741,17 @@ func runKubectl(t *testing.T, args ...string) {
 // applyToNamespace reads a manifest file, replaces all occurrences of "e2e-test-ns"
 // with the actual namespace, and applies it to the cluster.
 func applyToNamespace(t *testing.T, manifestPath, namespace string) {
+	applyManifestReplacing(t, manifestPath, namespace, "e2e-test-ns")
+}
+
+// applyManifestReplacing reads a manifest file, replaces placeholderNamespace with namespace, and applies it.
+func applyManifestReplacing(t *testing.T, manifestPath, namespace, placeholderNamespace string) {
 	content, err := os.ReadFile(manifestPath)
 	if err != nil {
 		t.Fatalf("failed to read manifest %s: %v", manifestPath, err)
 	}
 
-	// Replace all occurrences of e2e-test-ns with the actual namespace
-	modifiedContent := strings.ReplaceAll(string(content), "e2e-test-ns", namespace)
+	modifiedContent := strings.ReplaceAll(string(content), placeholderNamespace, namespace)
 
 	cmd := exec.CommandContext(context.Background(), "kubectl", "apply", "-f", "-")
 	cmd.Stdin = strings.NewReader(modifiedContent)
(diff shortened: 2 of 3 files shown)


david-martin:
Thanks for putting this together @HarshithaMS005

The main value I see here is as a regression test on the quickstart YAML files themselves (Gateway, HTTPRoutes, XBackends, XAccessPolicies), which I believe is worth having e.g. if someone breaks policy/e2e.yaml or the MCP server deployment, this catches it before users hit it.

One thing worth checking, are assertToolCallAllowed, assertMCPRequestAllowed, and initializeMCPWithCertsOptional reimplementing logic that already exists in checkToolCall and initializeMCPWithCerts?
Could these compose on top of the existing helpers with some small refactoring?
Not a blocker, but it might cut ~100 lines and keep things from drifting apart.

HarshithaMS005:
> Thanks for putting this together @HarshithaMS005
> 
> The main value I see here is as a regression test on the quickstart YAML files themselves (Gateway, HTTPRoutes, XBackends, XAccessPolicies), which I believe is worth having e.g. if someone breaks policy/e2e.yaml or the MCP server deployment, this catches it before users hit it.
> 
> One thing worth checking, are assertToolCallAllowed, assertMCPRequestAllowed, and initializeMCPWithCertsOptional reimplementing logic that already exists in checkToolCall and initializeMCPWithCerts? Could these compose on top of the existing helpers with some small refactoring? Not a blocker, but it might cut ~100 lines and keep things from drifting apart.

@david-martin Addressed — refactored the MCP helpers to compose on shared primitives (tryInitializeMCPWithCerts, doMCPRequest, checkMCPAllowed) instead of duplicating curl/parse/retry logic. Moved assertToolCallAllowed and assertMCPRequestAllowed into controller_test.go alongside the other MCP helpers. Test behaviour is unchanged.

david-martin:
/lgtm
/approve

This adds good regression coverage on the quickstart manifests (Gateway, HTTPRoutes, XBackends, XAccessPolicies). If someone breaks policy/e2e.yaml or the MCP server deployment, CI catches it before users hit it.

The refactoring to compose on shared primitives (`tryInitializeMCPWithCerts`, `doMCPRequest`, `checkMCPAllowed`) looks clean and addresses my earlier feedback.

If maintaining this becomes a burden as the quickstart evolves, we can move it out of per-PR CI and run it per-release instead, potentially with broader coverage (quickstart script execution, external MCP backends).

PR #287, opened 2026-05-24
fix: Build controller from source in quickstart to avoid stale image

### LINKED ISSUE #286: Quickstart broken: stale controller image doesn't recognize v1alpha1 policies
## Symptom

Running `make quickstart` (or `make quickstart-ollama`, `make quickstart-gemini`) results in all tool calls from the agent returning "Access to this tool is forbidden."

## Root Cause

The quickstart script (`site-src/guides/quickstart/run-quickstart.sh`) deploys the controller using the remote image `us-central1-docker.pkg.dev/k8s-staging-images/agentic-net/agentic-networking-controller:main`. This image is stale — it predates PR #271 which migrated the controller from watching `v0alpha0` to `v1alpha1` XAccessPolicies.

The quickstart policies are `v1alpha1`, but the stale controller image only recognizes `v0alpha0` authorization types, causing all tool calls to be denied.

The E2E tests do not have this problem because they build the controller from source via `build_and_load_controller_image()` in `dev/ci/lib.sh`.

## Workaround

Run `make dev-reload-controller` after quickstart completes.

## Fix

Modify `run-quickstart.sh` to build and load the controller image from source (the same way E2E tests do) instead of relying on the remote `:main` image. This ensures the quickstart always uses a controller that matches the policies in the repo.

Note: Go toolchain (Go 1.25+) is already a prerequisite for the quickstart (checked explicitly in the script), so building from source adds no new dependency for users.

A self-contained fix (building from source) is being submitted as a PR.

## CI Fix (Parallel Track)

A CI fix to ensure Google Cloud Build rebuilds the `:main` image on every merge to main is being investigated in parallel. This would make the remote image reliable again for other consumers. However, this may need more context on the Google Cloud Build process to ensure we're not stepping on any toes on the infra side.

rubambiza:
## Summary
- The quickstart was pulling a remote controller image (`:main` tag) that predates PR #271's v0alpha0→v1alpha1 migration, causing all tool calls to return "Access to this tool is forbidden"
- Fix: build the controller from source (Go toolchain is already a prerequisite), matching the pattern used by E2E tests
- CI fix to keep the remote `:main` image fresh is being investigated in parallel (#286)

## Test plan
- [x] Run `make quickstart-ollama` and verify tool calls succeed
- [x] Run `make quickstart` and verify tool calls succeed
- [x] Verify `make dev-reload-controller` still works as a manual fallback

Fixes #286

🤖 Assisted by [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/site-src/guides/quickstart/run-quickstart.sh
+++ b/site-src/guides/quickstart/run-quickstart.sh
@@ -34,6 +34,11 @@ GATEWAY_API_VERSION="v1.5.0"
 AGENT_UI_PORT="8081"
 AGENT_UI_URL="http://localhost:${AGENT_UI_PORT}/dev-ui/?app=mcp_agent"
 
+# Controller image settings (matches dev/ci/lib.sh)
+CONTROLLER_REGISTRY="us-central1-docker.pkg.dev/k8s-staging-images/agentic-net"
+CONTROLLER_IMAGE_NAME="agentic-networking-controller"
+CONTROLLER_TAG="main"
+
 # Default to HuggingFace, can be overridden with --ollama or --gemini flags
 USE_OLLAMA=false
 OLLAMA_BASE_URL="http://host.docker.internal:11434"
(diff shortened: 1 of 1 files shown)


david-martin:
/approve

running a quick test locally

PR #291, opened 2026-06-05
Document MCP params matcher context for XAccessPolicy

### LINKED ISSUE #275: Document context of v1alpha1's MCP params matchers
In MCP, "params" are method-specific. v1alpha1 XAccessPolicy's [`spec.rules.authorization.mcp.methods.params`](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/8331f16c5a53a6f6620e05dd5342b8a681ad89c8/api/v1alpha1/accesspolicy_types.go#L215) should document the context of the matcher relatively to each MCP method it can be used with.

E.g., an MCP `tools/call` request looks like this:

```json
{
  "method":"tools/call",
  "params":{
    "name":"get-sum",
    "arguments":{"a":2,"b":3},
    "_meta":{"progressToken":1}
  },
  "jsonrpc":"2.0",
  "id":1
}
```

So for `tools/call`, `xaccesspolicies.spec.rules.authorization.mcp.methods.params` matches against the `name` param of the MCP request.

In turn, a `resources/subscribe` request looks like this:

```json
{
  "method": "resources/subscribe",
  "params": {
    "uri": "file:///home/bob/config.yaml"
  },
  "jsonrpc": "2.0",
  "id": 2
}
```

Here, I imagine `xaccesspolicies.spec.rules.authorization.mcp.methods.params` matches against the `uri` param of the MCP request.

I think this requires documentation.

E.g.:

| MCP method              | XAccessPolicy param match |
|:------------------------|:-------------------------:|
| `prompts/get`           | `name`                    |
| `resources/subscribe`   | `uri`                     |
| `resources/unsubscribe` | `uri`                     |
| `resources/read`        | `uri`                     |
| `tools/call`            | `name`                    |

Or if we want to simplify it:

| MCP primitive | XAccessPolicy param match |
|:--------------|:-------------------------:|
| `prompts`     | `name`                    |
| `resources`   | `uri`                     |
| `tools`       | `name`                    |

HarshithaMS005:
**What type of PR is this?**
/kind documentation

In MCP, the JSON-RPC params object is method-specific, but v1alpha1 XAccessPolicy did not clearly document which MCP request field each value in [spec.rules.authorization.mcp.methods.params](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/8331f16c5a53a6f6620e05dd5342b8a681ad89c8/api/v1alpha1/accesspolicy_types.go#L215) is matched against. For example, tools/call uses params.name (the tool name), while resources/subscribe uses params.uri.
This PR documents that mapping so policy authors know what to put in params for each MCP method. 
It updates:
- API type comments on MCPMethod.Params and MCPMethodParam
- CRD OpenAPI descriptions (regenerated from the Go types)
- Site introduction docs with per-method tables and JSON-RPC examples
- The Tool Auth API proposal under v1alpha1 implementation status

Fixes #275 

```release-note
NONE
```

### DIFF
--- a/api/v1alpha1/accesspolicy_types.go
+++ b/api/v1alpha1/accesspolicy_types.go
@@ -245,16 +245,31 @@ type MCPMethod struct {
 	// +required
 	Name MCPMethodName `json:"name"`
 
-	// Params allows matching against specific arguments in the MCP request.
-	// Only valid for 'get', 'call', 'subscribe', 'unsubscribe', and 'read' methods.
-	// If empty or omitted, parameter-level allowlisting is not applied, meaning the method
-	// is authorized regardless of the arguments passed in the request.
+	// Params lists values to match against the "name" field in the MCP JSON-RPC request's
+	// "params" object. In the current prototype implementation, all supported methods
+	// (prompts/get, tools/call, resources/subscribe, resources/unsubscribe, and
+	// resources/read) match against "params.name".
+	//
+	// Note: In the MCP protocol, resources/subscribe, resources/unsubscribe, and
+	// resources/read use "params.uri" rather than "params.name". Matching against
+	// "params.uri" for resource methods is not yet implemented and will be addressed
+	// separately.
+	//
+	// For example, a tools/call request with params {"name":"get-sum","arguments":{"a":2,"b":3}}
+	// is matched by a param value of "get-sum".
+	//
+	// Only valid for prompts/get, tools/call, resources/subscribe, resources/unsubscribe,
+	// and resources/read. If empty or omitted, parameter-level allowlisting is not applied,
+	// meaning the method is authorized regardless of the arguments passed in the request.
 	// +optional
 	// +listType=set
 	// +kubebuilder:validation:MaxItems=10
 	Params []MCPMethodParam `json:"params,omitempty"`
 }
 
+// MCPMethodParam is a value to match against the "name" field in the MCP JSON-RPC
+// request's "params" object. See MCPMethod.Params for current matching behavior
+// and known limitations for resource methods.
 // +kubebuilder:validation:MaxLength=20
 type MCPMethodParam string
 
--- a/docs/proposals/0008-ToolAuthAPI.md
+++ b/docs/proposals/0008-ToolAuthAPI.md
@@ -23,6 +23,20 @@ A prototype implementation exists. **Use the implemented API below**, not the or
 | `serviceAccounts: ["ns/sa"]` shorthand | `source.type: ServiceAccount` + `serviceAccount: { name, namespace? }` |
 | Rule-level external auth | `spec.action: ExternalAuth` + `spec.externalAuth` |
 
+### MCP `params` matching (v1alpha1)
+
+In MCP JSON-RPC requests, the `params` object is method-specific. In `spec.rules[].authorization.mcp.methods[].params`, each listed value is matched against the `name` field in the MCP request's `params` object.
+
+| MCP method | Currently matched field |
+| --- | --- |
+| `prompts/get` | `name` |
+| `tools/call` | `name` |
+| `resources/subscribe` | `name` |
+| `resources/unsubscribe` | `name` |
+| `resources/read` | `name` |
+
+**Known limitation:** In the MCP protocol, resource methods use `params.uri` rather than `params.name`. The prototype implementation currently matches `params.name` for all methods, including resources. Matching against `params.uri` for resource methods is not yet implemented and will be addressed separately.
+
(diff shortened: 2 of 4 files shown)


api/v1alpha1/accesspolicy_types.go
```diff
@@ -245,16 +245,33 @@ type MCPMethod struct {
...
 	Name MCPMethodName `json:"name"`
 
-	// Params allows matching against specific arguments in the MCP request.
-	// Only valid for 'get', 'call', 'subscribe', 'unsubscribe', and 'read' methods.
-	// If empty or omitted, parameter-level allowlisting is not applied, meaning the method
-	// is authorized regardless of the arguments passed in the request.
+	// Params lists values to match against a specific field in the MCP JSON-RPC request's
+	// "params" object. MCP request params are method-specific; which field is matched
+	// depends on the MCP method in Name:
+	//
```
LiorLieberman: these are going into api comments. I am not sure we want such level of details on the api docs but maybe I am wrong. 

@guicassolato @david-martin what do you think? Is there an alternative better place way to document that you have in mind? Or maybe we we mention something short here (like what param means for every method) and something longer somewhere else?

LiorLieberman (commented):
Thank you @HarshithaMS005 ! Left one comment and asked Daivd and Gui's thoughts about it.

Also, we meet every Thu 8am pacific - would love to have you there and talk about ideas/work

api/v1alpha1/accesspolicy_types.go
david-martin: I'm in favour of keeping the verbose comments here. A few reasons:

1. These comments are the source of truth. They feed the CRD schema, `kubectl explain`, and the generated API reference. Users reaching for `kubectl explain` expect to find the detail there.

2. It's not either/or. Guides can reference and expand on this, but the API types should be self-documenting.

3. There's precedent in the ecosystem for this level of detail on struct fields:
   - [Core k8s `IngressRule.Host`](https://github.com/kubernetes/api/blob/v0.36.0/networking/v1/types.go#L400-L423) - ~24 lines covering matching semantics, RFC deviations, wildcard rules
   - [Gateway API `GatewaySpec.Addresses`](https://github.com/kubernetes-sigs/gateway-api/blob/v1.5.1/apis/v1/gateway_types.go#L247-L274) - ~15 lines of MUST/MAY implementation behavior
   - [Gateway API `HeaderMatchType`](https://github.com/kubernetes-sigs/gateway-api/blob/v1.5.1/apis/v1/httproute_types.go#L554-L568) - per-enum-value conformance levels inline, which is the closest parallel to the per-method breakdown here

guicassolato: I see no harm in keeping it in the API docs either at first, but what's the concern exactly @LiorLieberman?

You thinking about too much coupling to the _current_ MCP spec perhaps, and possibly use deliberately vague language instead for the sake of maintainability of our docs?

The part that goes right after this line – i.e., what MCP param is matched depending on the method – seems important to me. I would probably have a different opinion if the API required users to specify the match explicitly – e.g. `mcp.methods.params.name` and `mcp.methods.params.uri`, instead of more abstract form `mcp.methods.params: []string`.

david-martin:
Thanks for this. One concern: the docs say `resources/*` methods match against `uri`, but the implementation always matches `params.name` for all methods. See [L532-558 on main](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/main/pkg/translator/accesspolicy.go#L532-L558).

A policy author writing `params: ["file:///home/bob/config.yaml"]` for `resources/subscribe` (as the new docs suggest) would get a rule that silently never matches, since it checks `params.name` not `params.uri`.

Could we document what the code actually does today (always `params.name`) and note `uri` matching for resources as a gap to fix separately?

HarshithaMS005:
> Thanks for this. One concern: the docs say `resources/*` methods match against `uri`, but the implementation always matches `params.name` for all methods. See [L532-558 on main](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/main/pkg/translator/accesspolicy.go#L532-L558).
> 
> A policy author writing `params: ["file:///home/bob/config.yaml"]` for `resources/subscribe` (as the new docs suggest) would get a rule that silently never matches, since it checks `params.name` not `params.uri`.
> 
> Could we document what the code actually does today (always `params.name`) and note `uri` matching for resources as a gap to fix separately?

Thanks @david-martin — good catch. You're right: the translator always matches params.name (pkg/translator/accesspolicy.go L558), so documenting uri for resource methods would mislead policy authors.
Updated the docs to reflect current behavior (always params.name) and added a note that MCP resource methods use params.uri, which is a known implementation gap to address separately.

LiorLieberman:
/retest
/approve

Please fix the errors @HarshithaMS005 and ping me for lgtm. thanks!

HarshithaMS005:
@LiorLieberman E2E keeps failing with the same error on retest: quickstart-everything-mcp:main: not found (ImagePullBackOff). Controller image builds and deploys successfully. This appears to be a missing staging image, unrelated to this this PR.

PR #293, opened 2026-06-08
Fix: Treat explicit empty MCP methods as deny-by-default in RBAC

### LINKED ISSUE #217: AccessPolicy RBAC denies requests incorrectly

HarshithaMS005:
**What type of PR is this?**
/kind bug

AccessPolicy RBAC translation incorrectly treated an explicit empty MCP methods allowlist as allow-all, so policies intended to deny tool access instead granted it. When multiple Allow policies targeted the same backend or gateway, rules could also collide on name and overwrite each other in the merged RBAC config.
This PR fixes that by:
- Treating explicit mcp.methods: [] as deny-by-default (the rule contributes no allow permissions)
- Preserving allow-all behavior when methods is omitted
- Using unique RBAC policy keys ({namespace}/{policy-name}/{rule-name}) when merging multiple Allow policies on the same target
- Aligning merged evaluation with permit-overrides semantics: a request is allowed if any Allow policy permits it

Fixes #217 

```release-note
Fixed AccessPolicy RBAC so explicit empty `mcp.methods: []` denies tool access instead of allowing all MCP traffic.
```

### DIFF
--- a/api/v1alpha1/accesspolicy_types.go
+++ b/api/v1alpha1/accesspolicy_types.go
@@ -50,7 +50,7 @@ type AccessPolicySpec struct {
 	// 1. ExternalAuth runs before all other Allow policies.
 	// 2. If an ExternalAuth server denies the request, the request is denied.
 	// 3. If it allows the request, processing continues for all other allow policies for that target.
-	// 4. The request is allowed only if all allow policies allow it.
+	// 4. The request is allowed if any allow policy allows it.
 	// +required
 	Action AccessPolicyActionType `json:"action"`
 
@@ -209,7 +209,8 @@ type MCPAttributes struct {
 	// Methods is a list of specific MCP functional methods to match.
 	// If specified, only MCP requests with a method
 	// that matches one of these items will be authorized.
-	// If empty or omitted, no method-level allowlisting is applied, meaning all
+	// If set to an empty list, no MCP methods are permitted (deny-by-default).
+	// If omitted, no method-level allowlisting is applied, meaning all
 	// MCP methods (e.g., all tools, prompts, and resources) are permitted.
 	// +kubebuilder:validation:MaxItems=10
 	// +optional
--- a/k8s/crds/agentic.networking.x-k8s.io_xaccesspolicies.yaml
+++ b/k8s/crds/agentic.networking.x-k8s.io_xaccesspolicies.yaml
@@ -822,7 +822,7 @@ spec:
                   1. ExternalAuth runs before all other Allow policies.
                   2. If an ExternalAuth server denies the request, the request is denied.
                   3. If it allows the request, processing continues for all other allow policies for that target.
-                  4. The request is allowed only if all allow policies allow it.
+                  4. The request is allowed if any allow policy allows it.
                 enum:
                 - Allow
                 - ExternalAuth
@@ -1117,7 +1117,8 @@ spec:
                                 Methods is a list of specific MCP functional methods to match.
                                 If specified, only MCP requests with a method
                                 that matches one of these items will be authorized.
-                                If empty or omitted, no method-level allowlisting is applied, meaning all
+                                If set to an empty list, no MCP methods are permitted (deny-by-default).
+                                If omitted, no method-level allowlisting is applied, meaning all
                                 MCP methods (e.g., all tools, prompts, and resources) are permitted.
                               items:
                                 description: MCPMethod defines a specific MCP method
--- a/pkg/translator/accesspolicy.go
+++ b/pkg/translator/accesspolicy.go
@@ -180,57 +180,72 @@ func (t *Translator) mergeAllowPoliciesToRBAC(policies []*agenticv1alpha1.XAcces
 
 	for _, policy := range policies {
 		for _, rule := range policy.Spec.Rules {
-			policyName := rule.Name
-			source := t.ruleSourceToPrincipalName(policy.Namespace, rule.Source)
-
-			var principalIDs []*rbacconfigv3.Principal
-			if source != "" {
-				principalIDs = append(principalIDs, &rbacconfigv3.Principal{
-					Identifier: &rbacconfigv3.Principal_Authenticated_{
-						Authenticated: &rbacconfigv3.Principal_Authenticated{
-							PrincipalName: &matcherv3.StringMatcher{
-								MatchPattern: &matcherv3.StringMatcher_Exact{Exact: source},
-							},
-						},
-					},
-				})
+			rbacPolicy, ok := t.buildAllowRBACPolicyFromRule(policy.Namespace, rule)
+			if !ok {
+				continue
 			}
-			if len(principalIDs) == 0 {
-				principalIDs = []*rbacconfigv3.Principal{buildAnyPrincipal()}
-			}
-
-			rbacPolicy := &rbacconfigv3.Policy{
-				Principals: principalIDs,
-			}
-
-			if rule.Authorization != nil {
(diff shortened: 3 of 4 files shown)


LiorLieberman:
Hi @HarshithaMS005, thank you

whats the rationale you believe that this should be treated as deny-by-default if someone provides an empty set?

We have it in the spec that it should be allowed but open to your thoughts on why should we do differently. 

https://github.com/kubernetes-sigs/kube-agentic-networking/blob/main/api/v1alpha1/accesspolicy_types.go#L212-L218

david-martin:
Took me a while to understand the nuance of this change (and I may forget again and need to refresh my memory in future).
As I read it, trying to put some meaning on it, it's like saying 'heres your allow list, and its empty, so no one gets through'. So it's a zero length permission.

I wonder if the idea of a default deny should be more explicit.

I worry this subtlety could be lost in some implementations of controllers. Also, different impls may be in languages that serialize empty arrays in different ways, and lose the meaning.

HarshithaMS005:
> Hi @HarshithaMS005, thank you
> 
> whats the rationale you believe that this should be treated as deny-by-default if someone provides an empty set?
> 
> We have it in the spec that it should be allowed but open to your thoughts on why should we do differently.
> 
> https://github.com/kubernetes-sigs/kube-agentic-networking/blob/main/api/v1alpha1/accesspolicy_types.go#L212-L218

Hi @LiorLieberman — thanks for the question, this is a good point to clarify.
I'm not proposing a new semantic. This change aligns v1alpha1 behavior with the original AccessPolicy design and with [#217](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/217).
The current v1alpha1 comment at accesspolicy_types.go L212–218 treats empty the same as omitted. That looks like a documentation/implementation regression from v0alpha0, which explicitly states:

"If the tool list is empty, the rule denies access to all tools from Source."

The same intent is documented in [proposal 0008](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/main/docs/proposals/0008-ToolAuthAPI.md).

Rationale: Inline authorization is an allowlist.

methods: [] (explicitly set) → zero-length allowlist → allow nothing from this rule (deny-by-default baseline)
methods omitted → no MCP-level restriction for that rule
The bug was that the translator treated [] as allow-all via buildAnyPermission(), which breaks the #217 scenario where one policy establishes a deny-by-default baseline and another policy incrementally allows specific tools.

This PR updates the v1alpha1 comment and CRD to reflect that distinction. Happy to discuss making default-deny more explicit in the API if we want to reduce reliance on [] vs omitted across implementations

HarshithaMS005:
> Took me a while to understand the nuance of this change (and I may forget again and need to refresh my memory in future). As I read it, trying to put some meaning on it, it's like saying 'heres your allow list, and its empty, so no one gets through'. So it's a zero length permission.
> 
> I wonder if the idea of a default deny should be more explicit.
> 
> I worry this subtlety could be lost in some implementations of controllers. Also, different impls may be in languages that serialize empty arrays in different ways, and lose the meaning.

Thanks @david-martin , thats a helpful way to put it.
Agree — explicit methods: [] is a zero-length allowlist (“allow nothing from this rule”), and I share the concern that this is subtle across implementations/languages. This PR fixes the bug where empty allowlists were incorrectly treated as allow-all, matching v0alpha0 / #217. I’m open to a follow-up to make default-deny more explicit in the API so controllers don’t have to rely on [] vs omitted.

LiorLieberman:
>Took me a while to understand the nuance of this change (and I may forget again and need to refresh my memory in future).
>As I read it, trying to put some meaning on it, it's like saying 'heres your allow list, and its empty, so no one gets through'. So it's a zero length permission.
I wonder if the idea of a default deny should be more explicit.

>I worry this subtlety could be lost in some implementations of controllers. Also, different impls may be in languages that serialize empty arrays in different ways, and lose the meaning.

This brings significant challenges of usability and goes in a different direction that the prior art we are familiar with (e.g. core k8s networkpolicy v1 and istio authorizationpolicy.)

Imagine you want to deploy a policy that say "allow traffic from this source" - if you need to go and enumerate all mcp methods + any other request param and allow it there its not scalable or usable. when we introduce headers - you would also need to allowlist those?

therfore, usually if its omitted/empty - its treated as "match everything". 

For example - network policy ports 

```
spec:
  podSelector: { matchLabels: { app: backend } }
  ingress:
  - from:
    - podSelector: { matchLabels: { app: frontend } }
    # 'ports' is omitted here -> allows ALL ports
 ```
 
 and istio is another example
 
 ```
 spec:
  selector: { matchLabels: { app: backend } }
  action: ALLOW
  rules:
  - from:
    - source: { principals: ["cluster.local/ns/default/sa/adk-agent"] }
    # 'to' section (paths/methods) is omitted -> allows ALL paths and methods
 ```
 
 >methods: [] (explicitly set) → zero-length allowlist → allow nothing from this rule (deny-by-default baseline)
methods omitted → no MCP-level restriction for that rule

@HarshithaMS005 the challenge here is that you can not easily distinguish in go between empty and omitted unless you turn this into a pointer. I see your PR does **not** change the field to a pointer (its good cause we dont want that) how can it distinguish between empty and ommited?

Thanks!

HarshithaMS005:
Thanks @LiorLieberman — really helpful framing.

**On "allow traffic from this source"**
Agree this must stay simple and match prior art for the common case. That case is already covered without enumerating MCP methods:
- omit `authorization` on the rule -> allow all from that source (same idea as omitting Istio `to`, or NetPol `ports`)
- or `authorization.type: Inline` with `mcp` / `mcp.methods` omitted -> no MCP-level restriction

`methods: []` is not intended for "allow from source"; it's for the #217 incremental-allow pattern (deny baseline + another policy allowing specific tools).

**On empty vs omitted in Go (without pointers)**
In this implementation we distinguish via slice nil-ness after API unmarshaling:
- `methods` omitted -> `Methods == nil` -> no method restriction
- `methods: []` present -> `Methods != nil && len(Methods) == 0` -> explicit zero-length allowlist
See `translateMCPToRBACPermissions` (`Methods != nil && len(Methods) == 0`).

I agree this is fragile across languages/implementations if clients normalise `nil` and `[]` — same concern @david-martin raised. A more explicit API for default-deny would be better long-term.

**On prior art (NetPol / Istio)**
Understood that omitted constraints usually mean "match everything." The empty=deny behaviour comes from v0alpha0 / proposal 0008 / #217, not from trying to make the common "allow from source" path harder.

**Question:**
Should v1alpha1 align with NetPol/Istio (empty **and** omitted = allow all), and handle #217 via a future explicit deny mechanism? Or keep empty `methods: []` as deny-by-default for #217 now, with a follow-up to make that explicit in the API?

Happy to adjust the PR based on that decision. The RBAC merge-key fix (multiple policies overwriting each other) is independent and can land either way.

david-martin:
@HarshithaMS005 Checking in, is this still something you are interested in landing?

I think there's still value in firming up on the behaviour for allow all vs allow none

PR #294, opened 2026-06-09
Add MCPGatewayExtension proposal

david-martin:
Proposes `MCPGatewayExtension`, a CRD using the policy attachment pattern to signal a Gateway implementation to enable MCP request body parsing on a targeted listener.

Key points:
- Listener-level targeting via `sectionName`, with fallback to all listeners when omitted
- Discusses alternatives (gateway-level only, per-route targeting)
- Notes the incoming MCP spec RC (2026-07-28) `Mcp-Method`/`Mcp-Name` headers, and the transition period where body parsing is still needed
- Deliberately leaves extracted fields and implementation details open

Replaces the earlier dual-approach proposal (MCPGatewayExtension vs Gateway `mcpMode` field) with a single focused design.

### DIFF
--- a/docs/proposals/XXXX-MCPGatewayExtension.md
+++ b/docs/proposals/XXXX-MCPGatewayExtension.md
@@ -1,0 +1,85 @@
+Date: 9th June 2026<br/>
+Authors: david-martin<br/>
+Status: Provisional<br/>
+
+# MCPGatewayExtension
+
+## Summary
+
+MCP-aware features like `XAccessPolicy` tool authorization require the Gateway to parse MCP protocol messages. There is no standard mechanism to enable this. This proposal introduces `MCPGatewayExtension`, a CRD that targets a Gateway listener and signals the implementation to enable MCP request body parsing on that listener.
+
+## Non-Goals
+
+- Defining what MCP parsing means at the implementation level (buffering, streaming, external processing).
+- Authorization or access control policy (`XAccessPolicy`).
+- Backend registration or routing (`XBackend`, `HTTPRoute`).
+- Locking down which fields are extracted from the request body. The set of extracted information will evolve alongside the MCP specification.
+- Implementation-specific concerns: observability, auth, rate limiting.
+
+## Use Cases & Motivation
+
+A platform operator deploys a Gateway and wants `XAccessPolicy` rules to apply to MCP tool calls. Without a signal to the Gateway that MCP parsing should be active on a listener, MCP traffic is opaque HTTP and tool-level policies have no effect.
+
+## Design
+
+### CRD
+
+`MCPGatewayExtension` uses the Gateway API policy attachment pattern. `sectionName` targets a specific listener. When omitted, parsing is enabled on all listeners.
+
+```yaml
+apiVersion: agentic.networking.x-k8s.io/v1alpha1
+kind: MCPGatewayExtension
+metadata:
+  name: enable-mcp
+  namespace: gateway-system
+spec:
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: Gateway
+    name: prod-gateway
+    sectionName: mcp-listener
+```
+
+```go
+type MCPGatewayExtensionSpec struct {
+    // TargetRef identifies the Gateway (and optionally a specific listener)
+    // on which to enable MCP parsing.
+    // +required
+    TargetRef gwapiv1.LocalPolicyTargetReferenceWithSectionName `json:"targetRef"`
+}
+```
+
+What happens when a listener is targeted is implementation-specific. In the Istio/Envoy case, this could mean configuring an ext-proc filter for the listener's port. The outcome is that the Gateway can parse MCP JSON-RPC request bodies and surface extracted information (e.g., as headers or dynamic metadata) for routing, access control, or observability.
+
+### Why Listener-Level Targeting?
+
+Gateways typically serve multiple listeners: HTTPS on 443, admin endpoints, other. Parsing request bodies on listeners that don't carry MCP traffic adds overhead for no benefit. `sectionName` provides per-listener granularity using the same mechanism Gateway API policies already use.
+
+#### Alternatives
+
+**Gateway-level only (no per-listener targeting):** Simpler, but forces operators to either dedicate entire Gateways to MCP traffic or accept body parsing overhead on all listeners. Many deployments will share a Gateway across MCP and non-MCP traffic.
+
+**Per-route targeting (HTTPRoute):** Body parsing is typically configured at the infrastructure layer (e.g., Envoy filter chains operate at the listener/port level). Per-route targeting would also couple parsing configuration to routing configuration, complicating lifecycle management.
+
+### What Gets Extracted
+
+The MCP specification [release candidate (2026-07-28)](https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/) requires `Mcp-Method` and `Mcp-Name` headers on every Streamable HTTP request, mirroring information from the JSON-RPC body. This allows some method-level and tool-level policy enforcement from headers alone.
+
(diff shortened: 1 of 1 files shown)


LiorLieberman:
@david-martin This is great! do you need anything to drive this forward?

david-martin:
> do you need anything to drive this forward?

I've updated this to focus on the CRD mechanism for now, and point out some changes in incoming mcp spec that impacts this.
This CRD acts as a gate for if a gateway will accept/program (TBD) an AccessPolicy.
moving out of draft.

docs/proposals/XXXX-MCPGatewayExtension.md
```diff
@@ -0,0 +1,85 @@
...
+
+- Defining what MCP parsing means at the implementation level (buffering, streaming, external processing).
+- Authorization or access control policy (`XAccessPolicy`).
+- Backend registration or routing (`XBackend`, `HTTPRoute`).
+- Locking down which fields are extracted from the request body. The set of extracted information will evolve alongside the MCP specification.
+- Implementation-specific concerns: observability, auth, rate limiting.
+
+## Use Cases & Motivation
+
+A platform operator deploys a Gateway and wants `XAccessPolicy` rules to apply to MCP tool calls. Without a signal to the Gateway that MCP parsing should be active on a listener, MCP traffic is opaque HTTP and tool-level policies have no effect.
```
LiorLieberman, line 21:
```suggestion
A platform operator deploys a Gateway and wants `XAccessPolicy` rules to apply to MCP tool calls. In some gateway implementations, without a signal to the Gateway that MCP parsing should be active on a listener, MCP traffic is opaque HTTP and tool-level policies have no effect.
```

docs/proposals/XXXX-MCPGatewayExtension.md
```diff
@@ -0,0 +1,85 @@
...
+```go
+type MCPGatewayExtensionSpec struct {
+    // TargetRef identifies the Gateway (and optionally a specific listener)
+    // on which to enable MCP parsing.
+    // +required
+    TargetRef gwapiv1.LocalPolicyTargetReferenceWithSectionName `json:"targetRef"`
+}
+```
+
+What happens when a listener is targeted is implementation-specific. In the Istio/Envoy case, this could mean configuring an ext-proc filter for the listener's port. The outcome is that the Gateway can parse MCP JSON-RPC request bodies and surface extracted information (e.g., as headers or dynamic metadata) for routing, access control, or observability.
```
LiorLieberman, line 52: >What happens when a listener is targeted is implementation-specific.

Not sure I understand. Probably phrasing can be improved. Are you answering the question of what happens when some implementations does not support listener targeting OR when some implementations do not need listener targeting? 

Can you calrify?

docs/proposals/XXXX-MCPGatewayExtension.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Why Listener-Level Targeting?
+
+Gateways typically serve multiple listeners: HTTPS on 443, admin endpoints, other. Parsing request bodies on listeners that don't carry MCP traffic adds overhead for no benefit. `sectionName` provides per-listener granularity using the same mechanism Gateway API policies already use.
+
+#### Alternatives
+
+**Gateway-level only (no per-listener targeting):** Simpler, but forces operators to either dedicate entire Gateways to MCP traffic or accept body parsing overhead on all listeners. Many deployments will share a Gateway across MCP and non-MCP traffic.
```
LiorLieberman, line 60:
```suggestion
**Gateway-level only (no per-listener targeting):** Simpler, but forces operators to either dedicate entire Gateways to MCP traffic or accept body parsing overhead on all listeners. Many deployments will share a Gateway across MCP and non-MCP traffic. Additionally, we the proposed architecture supports both options (e.g when `sectionName` is not specified). We can set `sectionName` to be "Extended" conformance and avoid requiring all implementations to implement listener targeting if they don't wish to do it
```

docs/proposals/XXXX-MCPGatewayExtension.md
```diff
@@ -0,0 +1,85 @@
...
+
+### What Gets Extracted
+
+The MCP specification [release candidate (2026-07-28)](https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/) requires `Mcp-Method` and `Mcp-Name` headers on every Streamable HTTP request, mirroring information from the JSON-RPC body. This allows some method-level and tool-level policy enforcement from headers alone.
+
+While the new spec propagates through MCP client and server implementations, there will be a need to continue parsing the MCP method and tool name from the request body. In addition, there may be other request body information that implementations need to extract: tool call arguments, client metadata in `_meta`, or fields introduced by future spec revisions. What to extract and how to surface it is deliberately left open. `MCPGatewayExtension` signals that the Gateway *may* parse the body; what it does with the contents is implementation-defined.
```
LiorLieberman, line 68: >there will be a need to continue parsing the MCP method and tool name from the request body.

If those are soon to be available in headers, why would we continue to have this need?

rest of the argument lgtm

docs/proposals/XXXX-MCPGatewayExtension.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Status
+
+The resource carries its own status conditions (`Accepted`, `Programmed`), independent of the Gateway's status.
+
+### Constraints
+
+- One `MCPGatewayExtension` per listener. Conflicts are rejected and reported in status.
+- Implementations MUST NOT enforce MCP-aware policies unless parsing has been explicitly enabled.
```
LiorLieberman, line 77: please suggest how to handle cases of mcp-aware policies without MCPGatewayExtension. 

Or alternatively change MUST NOT to SHOULD NOT imo.  ( I think agentgateway are doing the parsing implicitly but I am not sure)

docs/proposals/XXXX-MCPGatewayExtension.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Extensibility
+
+The spec is minimal. Implementations may need buffer sizes, timeouts, session storage, TLS configuration. These can be added as optional fields without touching the Gateway resource, which is the main advantage of a separate CRD over modifying the Gateway spec directly.
+
+## Support Requirements
+
+An implementation that supports MCP-aware features MUST support `MCPGatewayExtension` or an equivalent mechanism that provides a clear, observable signal that MCP parsing is active on a given Gateway or listener.
```
LiorLieberman, line 85: >or an equivalent mechanism that provides a clear

The above is pretty vague. We can not say MUST do something but be leave it vague. Other propose something a little more clear or remove the MUST here and say that implementations that support it without it is undefined or up to implementation specific behavior or something

LiorLieberman (commented):
Thanks @david-martin !

sjberman:
Would attaching an XAccessPolicy that wants to parse MCP not be enough to signal to the Gateway that it needs to be MCP aware?

Just not sure why an additional resource is necessary if the Policy already tells the Gateway what it wants.

david-martin:
@sjberman 

> Would attaching an XAccessPolicy that wants to parse MCP not be enough to signal to the Gateway that it needs to be MCP aware?

Technically, that could work.
The reconcile logic for XAccessPolicy could ensure the gateway is configured to accept/parse/route/whatever MCP traffic once at least one XAccessPolicy exists.

What would the behaviour be when all XAccessPolicy resources are deleted?
Would the gateway be reconfigured to remove the MCP stuff?

> Just not sure why an additional resource is necessary if the Policy already tells the Gateway what it wants.

The resource is coming from an implementation that adds other configuration to the gateway around MCP traffic i.e. MCP tools/prompt/resource federation from upstream MCP servers with varying auth and protocol versions.
So it has a greater meaning and carries with it other config in practice.

I wonder if this should be more explicit, and change the proposal accordingly, with examples,
or revise the approach to be more about just enabling what's needed to be able to do XAccessPolicy things.

sjberman:
@david-martin 

> What would the behaviour be when all XAccessPolicy resources are deleted?
Would the gateway be reconfigured to remove the MCP stuff?

In theory, yes. But this is also coming from my lack of knowledge of what the burden is of reconfiguring the Gateway to add or remove the extra MCP config. This also could be implementation specific. Maybe for some, it requires a full reload and downtime. For others, it might not be noticeable at all.

PR #295, opened 2026-06-11
Define scripts and doc for release.

bexxmodd:
**What type of PR is this?**
/kind documentation

shell script to generate installation yaml which should be attached as a release artifact

```release-note
NONE
```

### DIFF
--- a/dev/ci/lib.sh
+++ b/dev/ci/lib.sh
@@ -195,7 +195,9 @@ install_crds() {
   kubectl apply --server-side -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.5.1/standard-install.yaml
 
   header "Installing Project CRDs"
-  kubectl apply -f k8s/crds/
+  for file in k8s/crds/agentic.networking.x-k8s.io_*.yaml; do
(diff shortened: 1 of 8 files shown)


dev/ci/lib.sh
```diff
@@ -195,7 +195,9 @@ install_crds() {
   kubectl apply --server-side -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.5.1/standard-install.yaml
 
   header "Installing Project CRDs"
-  kubectl apply -f k8s/crds/
+  for file in k8s/crds/agentic.networking.x-k8s.io_*.yaml; do
```
LiorLieberman, line 198: nit: why. this change is important?

LiorLieberman (commented):
/lgtm
/approve
/hold

dev/ci/lib.sh
bexxmodd: I added kustomization.yaml in the same directory that is used by build script, because of that this script (lib.sh) was trying to install kustomization.yaml as a CRD and was failing ci. This one just skips files that are not crd specific.

PR #296, opened 2026-06-12
add goreleaser config and build workflow on each PR.

bexxmodd:
**What type of PR is this?**
/kind documentation

This makes sure that controller binary builds correctly for each PR (in the future the same config will be used to automate the release #255 )

```release-note
NONE
```

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -45,7 +45,8 @@ release:
   header: |
     ## Installation
 
-    To install this release, refer to the instructions in the README.md.
+
+    To install this release, refer to the instructions in [README.md](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/v{{.Version}}/README.md).
   # extra_files:
   # - glob: ./release/*.yaml
 
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -46,7 +46,8 @@ release:
     ## Installation
 
 
-    To install this release, refer to the instructions in [README.md](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/v{{.Version}}/README.md).
+    To install this release, refer to the instructions in 
+    [README.md](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/v{{.Version}}/README.md).
   # extra_files:
   # - glob: ./release/*.yaml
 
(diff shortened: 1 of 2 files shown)


.goreleaser.yaml
```diff
@@ -0,0 +1,50 @@
...
+  name_template: 'checksums.txt'
+snapshot:
+  version_template: "{{ incpatch .Version }}-next"
+release:
+  draft: true
+  name_template: "Release-v{{.Version}}"
+  header: |
+    ## Installation
+
+    To install this release, refer to the instructions in the README.md.
```
LiorLieberman: should we add something like

https://github.com/kubernetes-sigs/kube-agentic-networking/blob/v{{.Version}}/README.md

instead of README.md?

.goreleaser.yaml
```diff
@@ -0,0 +1,50 @@
...
+  extra_files:
+    - glob: ./release/*.yaml
```
LiorLieberman: whats the purpose of this? 

and also do we need/want `changelog` config here?

bexxmodd: This will be used for release workflow. To combine built binary and installation manifests and uploads them as assets to the GitHub Release draft, right alongside the compiled controller binaries and checksums.txt.

I can remove it for now and add back when I propose the release related PR.

.goreleaser.yaml
bexxmodd: IMO release documentation should not vary based on version. We can define installation guide in a separate doc and keep it up-do-date on main.

.goreleaser.yaml
bexxmodd: > and also do we need/want changelog config here?

We'll be doing (partially) manual curation of the changelog so no. Added explicitly for it to being disabled.

.goreleaser.yaml
LiorLieberman: but the guide may change between versions no? e.g at version 0.1 you mush use tool X to generate certification, and version 0.2 it is changed. 

Usually this is solved with versioned docs (which kinda do the same thing - docs per version/bracnh) but its probably an overkill to add now.

bexxmodd: Oh yes, you're right. Let's have README versioned here for now until we have versioned docs.

LiorLieberman (approved):
Thanks @bexxmodd !
/approve
/lgtm

/hold until checks and workflow pass - feel free to unhold then

Issue #243, opened 2026-04-17
Proposal: receipt-based admission control for agent-produced manifests


tomjwxf:
Hi maintainers, thanks for the work on kube-agentic-networking. The scope (identity, authorization, auditability for cross-boundary agent traffic inside Kubernetes) is the right shape for the next several years of platform engineering.

Filing this as a proposal for a concrete extension point: **receipt-based admission control for agent-produced manifests**. An AI agent that runs `kubectl apply`, modifies a CRD, or pushes a Helm release through tool calls should leave behind a signed decision-receipt chain, and the admission controller should verify that chain before admitting the manifest.

This complements the authorization-policy and agent-identity surfaces you already list in the charter: identity proves WHO the agent is; authorization proves the call is permitted; a receipt chain proves WHAT was authorized at each step of the agent session that produced the manifest.

## The pattern

An AI agent in a dev environment runs a session that ends with a new Deployment manifest. Every tool call it made during that session (file reads, build steps, kubectl diff) produced a signed decision receipt. The chain ends at the `kubectl apply` step, which submits the manifest to the cluster.

An admission webhook (`ValidatingAdmissionWebhook` or a Gateway API Policy Attachment) receives the submission plus a reference to the receipt chain, and runs:

1. Fetch the referenced receipt chain (from an OCI registry, HTTPS URL, or Sigstore Rekor entry).
2. Verify each receipt's Ed25519 signature against the declared supervisor-hook identity.
3. Walk the hash chain from genesis to the `kubectl apply` step; reject on any break.
4. Check that every receipt's `policy_digest` matches a cluster-approved policy.
5. Check that the `kubectl apply` receipt's decision is `allow` under a policy the cluster allows for manifest submissions.
6. Admit or deny the manifest based on the combined result.

## Where this fits in the existing charter

From the README:

> Define authorization policies that control which agents can communicate with other agents, tools, and LLMs at a granular level (e.g., specific MCP tool functions)

Receipts make those policy decisions verifiable after the fact. A cluster operator can answer "which policy allowed this manifest to be admitted?" by walking the receipt chain, without trusting the agent that produced it or the operator hosting the agent.

> Integrate AI safety and security extension points to support external authentication, authorization, and policy enforcement decisions

Receipts are the evidence artifact of those decisions. An admission controller that takes a receipt chain as input is a concrete shape for the "external enforcement decision" extension point.

[...]

PR #298, opened 2026-06-12
conformance: add receipt admission vectors

Rul1an:
Adds envelope-neutral vectors for the receipt-based admission discussion in #243. The vectors pin four admission concerns: chain-reference resolution, manifest binding, declared-vs-observed outcome, and canonicalization. They intentionally do not define the final Kubernetes object canonicalization profile; they make that open question explicit. Expected outcomes are reproducible from the bytes in `vectors.json` alone.

### DIFF
--- a/conformance/resources/receipt-admission/v0/README.md
+++ b/conformance/resources/receipt-admission/v0/README.md
@@ -1,0 +1,62 @@
+# Receipt admission conformance vectors (v0)
+
+Envelope-neutral conformance vectors for the receipt-based admission-control discussion in
+[#243](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/243). They pin four admission
+concerns, in the order a controller evaluates them, and every expected outcome is reproducible from the
+bytes in `vectors.json` alone.
+
+- **resolution** — the chain reference is fetched before anything is verified. A present but unresolvable
+  reference is rejected, because a committed digest in the annotation is not sufficient on its own.
+- **binding** — the controller recomputes the manifest digest from the object it actually received and
+  compares it to the chain's committed value, so a chain produced for one object cannot admit a
+  different one.
+- **declared-vs-observed** — for a bound chain, an independent observation of the effect either agrees
+  with, diverges from, or simply does not see what the receipt declares. A receipt is the agent
+  session's own account, so a missing observation is inconclusive, never confirmation.
+- **canonicalization** — the same object content under reordered keys and added whitespace must produce
+  one digest, so the profile (not a coincidence of two serializers) is what makes recompute
+  interoperable.
+
+These vectors deliberately do not define the final canonicalization profile for Kubernetes objects;
+they make that open question explicit and stay reproducible independent of it.
+
+## Format
+
+`vectors.json` carries a machine-readable canonicalization block and a list of vectors. A second
+implementation reproduces every verdict from these fields without any shared code:
+
+```json
+"canonicalization": {
+  "profile": "jcs-json-v1",
+  "description": "RFC 8785 (JCS) over the JSON object, restricted here to strings, objects, and small integers, so RFC 8785 number formatting is not exercised; a full implementation is required for general objects",
+  "hash": "sha256",
+  "digest_encoding": "hex",
+  "digest_prefix": "sha256:"
+}
+```
+
+To recompute a manifest digest: canonicalize the object under the named profile, hash it with the named
+hash, encode per `digest_encoding`, and prefix with `digest_prefix`. The committed digest is checked by
+recomputation, not trusted as supplied.
+
+## Verdicts
+
+- `resolution`: `unresolvable` (reason `chain_reference_unresolvable_digest_alone_insufficient`).
+- `binding`: `bound` | `digest_mismatch` | `missing_commitment`.
+- `grounding` (bound chains): `confirmed` | `contradicted` | `unobserved`, each with a `grounding_reason`.
+- `canonicalization`: `all_variants_match` | `variant_mismatch`.
+
+The required failure case is `g2`/`g3`: a cryptographically valid, bound chain whose independent observed
+effect diverges from the declaration. Vector `g5` is the case where a controller-signed attestation
+covers the submitted object but the effect is unobserved; it carries the distinct reason
+`object_attestation_not_effect_observation`, so attesting the object is never read as observing the
+effect, and it stays `unobserved`.
+
+## Scope
+
+Signature and chain-integrity verification is the emitter side and is assumed valid here; these vectors
+isolate resolution, binding, declared-vs-observed, and canonicalization. They are recompute-only and make
+no claim about any implementation. The canonical profile id `jcs-json-v1` is a placeholder for these
+fixtures; the production Kubernetes object profile (YAML/JSON normalization, defaulting, field ordering)
+is the open question for the proposal, and the committing and recomputing sides must share one named
+profile or recompute is not deterministic across implementations.
--- a/conformance/resources/receipt-admission/v0/README.md
+++ b/conformance/resources/receipt-admission/v0/README.md
@@ -1,7 +1,7 @@
 # Receipt admission conformance vectors (v0)
 
 Envelope-neutral conformance vectors for the receipt-based admission-control discussion in
-[#243](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/243). They pin four admission
+[#243](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/243). They pin five admission
 concerns, in the order a controller evaluates them, and every expected outcome is reproducible from the
 bytes in `vectors.json` alone.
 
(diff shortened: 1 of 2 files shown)


aeoess:
Ran the receipt-admission v0 vectors through an independent implementation, recomputing from `vectors.json` alone (rfc8785 + SHA-256, `sha256:` hex prefix, no shared code with this PR). All ten reproduce.

- binding (b1, b2, b3): recomputing the manifest digest from `admitted_object` gives `bound` on the match, `digest_mismatch` against the divergent commitment, and `missing_commitment` when none is supplied.
- canonicalization (c1): the reordered-keys variant and the whitespace variant both canonicalize to `sha256:7229caa8...41ff0`, collapsing to one digest under the profile.
- declared-vs-observed (g1 to g5): all bound; `confirmed` when the observation matches, `contradicted` when effect or target diverges, `unobserved` when there is no independent observation. g5 stays `unobserved` with `object_attestation_not_effect_observation`, so the controller attesting the object is not read as observing the effect.
- resolution (r1): the digest itself recomputes, but the chain reference is unresolvable, so it is `unresolvable` on the rule that a committed digest is not sufficient alone.

This is under the fixture's restricted `jcs-json-v1` profile (strings, objects, small integers; RFC 8785 number formatting not exercised), matching the README caveat. The production Kubernetes object profile (defaulting, field ordering, YAML/JSON normalization) is the open question these vectors leave open, so the agreement here is on the binding and canonicalization logic, not on a final object profile.

On the name: `receipt-admission.k8s-object.v0` works on our side. Once it lands in the proposal text we will pin our runner to that id so the two implementations cannot drift silently.

Rul1an:
Thanks for taking the time to recompute these independently, that's exactly the bar I hoped the vectors would clear. All ten reproducing from `vectors.json` alone, with no shared code, is the whole point: the bytes are meant to be the contract, not any one implementation. Good to see the rfc8785 + SHA-256 path lands cleanly on your side too.

The piece these deliberately leave open is the Kubernetes object canonicalization profile, so that question stays with #243 rather than being fixed here. Happy to adjust the vectors once the discussion there settles on one.

Ready for review whenever a kubernetes-sigs member has a moment to `/ok-to-test`.

aeoess:
The independent implementation reproduces the new stage family on this PR, and the full v0 set, from vectors.json alone. Pinned to head `4055d9f173e0bcd5d0b0a531106c9fd929a0a704`.

Method. Verdicts are derived, not echoed. For each stage vector the harness reads the commitment stage and the required stage, looks up the kind from the closed `stage_to_kind_to_verb` table, and applies the verb the table fixes for that stage. Object stages verify by recompute (JCS canonicalization then SHA-256, `sha256:` hex). The effect stage verifies by compare. The expected verdict in the file is never read into the decision.

Stage family, all seven reproduced:

- `st1` agent-declared and `st2` controller-observed-object: object kind, same stage, the recompute equals the commitment, so `bound`. `st2` recomputes the defaulted object (with `progressDeadlineSeconds`) to `sha256:09ba9cedc033ff71103dcd7efda1f3be9bb6cd2889a34360a87320cec2fe51ba`.
- `st4` controller-observed-object, same stage, with a changed `replicas` count: the recompute diverges, so `digest_mismatch` (tamper).
- `st3` commitment at agent-declared, required at controller-observed-object: object kind, different stage. The mismatch is the pipeline (defaulting), not tamper, so `cross_stage_digest_mismatch`. The agent-declared digest `sha256:7229caa820ff06fe2efd4bacd805b0f75891bd66297d234c450d9d30a3a41ff0` does not equal the controller-observed recompute, and the verdict follows from the stage difference, not from the digest divergence alone.
- `st5` (effect commitment, object required) and `st6` (object commitment, effect required): the required and commitment stages are different kinds, so `stage_kind_mismatch`, decided before any recompute or compare.
- `st7` observed-effect with a matching observation: `confirmed`.

Full set. The v0 families reproduce the same way, derived from the vector fields: resolution `unresolvable` (chain present, not resolvable); binding `bound`, `digest_mismatch`, `missing_commitment`; declared-vs-observed `confirmed`, two `contradicted`, two `unobserved` including the controller-object-attestation case; canonicalization `all_variants_match`, both raw variants collapsing to `sha256:7229caa820ff06fe2efd4bacd805b0f75891bd66297d234c450d9d30a3a41ff0`.

Seventeen vectors, every verdict reproduced. The stage-to-kind-to-verb table holds: the kind is fixed per stage, the verb derives from the stage, and the cross-stage and cross-kind rejects land where the table says they should.

The profile pin stays `jcs-json-v1`, matching the file, to be re-pinned when the production Kubernetes object profile is named.

Rul1an:
Seventeen verdicts derived independently from vectors.json, the stage family and all, with nothing shared between the two runners: that is the contract working as intended. What matters most to me landing on your side is the st3 reading. Your runner calls cross_stage_digest_mismatch off the stage difference, not off the digest divergence on its own, and holds it apart from the st4 same-stage tamper. Keeping those two distinct is the point of the whole family: defaulting is the pipeline, not an attack, and a suite that blurs them would quietly let a controller-observed recompute stand in for an agent-declared commitment. st5 and st6 rejecting on the stage kind before any verb runs is the same discipline from the other direction.

With the table reproduced both ways, the next move sits in #243: naming receipt-admission.k8s-object.v0 and its stage field in the proposal text as normative, the closed enum and the per-value kind, so both runners pin to an id rather than a shared convention. The production Kubernetes object profile stays the genuinely open part there, defaulting and field ordering and YAML/JSON normalization included; the stage family only fixes the verification contract a named profile has to satisfy, never the profile itself. Once the name lands I will move the vectors onto it and you can re-pin in step.

aeoess:
Staged the cross-stage negatives on our side for you to pull, same bar as the stage set: every verdict reproduces from the file alone, self-checked 5/5. They cover the directions #298 does not yet exercise.

- The reverse of `st3`: a commitment at controller-observed-object checked at agent-declared. Object kind on both sides, stages differ, so `cross_stage_digest_mismatch`.
- Stage-kind both ways anchored on agent-declared, since `st5` and `st6` both anchor on controller-observed-object: `stage_kind_mismatch`.
- `unknown_stage` from the required side and the commitment side, which the enum defines but no current vector exercises.

File and README: https://github.com/aeoess/aps-conformance-suite/tree/main/fixtures/k8s-receipt-admission-stage-negatives
Raw: https://raw.githubusercontent.com/aeoess/aps-conformance-suite/main/fixtures/k8s-receipt-admission-stage-negatives/stage-cross-negatives.candidate.json

One case I did not author, because the verdict is a table decision rather than something to assert unilaterally. A cross-stage object check where the two object stages produce the same digest, no defaulting between agent-declared and controller-observed-object. The stage difference points at `cross_stage_digest_mismatch`, but the equal digests read like `bound`, so the table has to say which wins. My read is to reject as cross-stage: a commitment carries the stage it was computed at, equal digests across different stages is a coincidence rather than a binding, and letting it pass as `bound` makes the stage field stop being load-bearing. The opposite read is defensible too, that an unchanged object is a no-op pipeline and the commitment still describes it. It is your enum, so I would rather raise it than encode it.

Rul1an:
Calling that case out instead of encoding it was the right move, and my read lands where yours does: a cross-stage object check rejects even when the digests are equal. The stage a commitment was computed at is part of the binding, not a label on it, so equal bytes across `agent-declared` and `controller-observed-object` is a coincidence rather than a binding. If we let it pass as `bound`, whether a controller-observed commitment satisfies an agent-declared requirement would hinge on whether defaulting happened to touch that object, and the stage field stops being load-bearing exactly where you cannot see it.

Where I would not reuse `cross_stage_digest_mismatch`: that name is false when the digests match. So this is a new sibling verdict, `cross_stage_match_rejected`, kept distinct from the divergent case. A divergent digest stays pipeline/defaulting; an equal digest is the coincidence reject; neither is `bound`, and `bound` keeps requiring same stage. That also leaves a controller free to treat a no-op pipeline leniently as policy, without the conformance verdict pre-blessing it.

I will add the same-digest vector under that verdict and pull your three cross-stage negatives into this PR, recomputing each from the file before they land so the two runners stay independent. The enum name and the per-stage kind move into #243 as normative next, and you can re-pin once the id is in the proposal text. The production object profile stays the open part there.

LiorLieberman:
Hi - would love one of your to share a little more on the PR + issue/proposal in our weekly meetings. At this point I am not really sure what is it or why we need it. 

I left a comment on #243 with the meeting cadence and links. 

Holding until we get more clarity on this
/hold

Rul1an:
Fair hold, and a reasonable thing to ask. Let me narrow what this PR actually is, because it is much smaller than the proposal thread around it.

#298 is not an admission controller and does not ask the SIG to adopt one. It is a small, envelope-neutral set of conformance vectors that pin one boundary: when a check binds a decision record to an object, the digest is recomputed from the object actually received under one named canonicalization, and "the decision was recorded" stays separate from "the effect was observed." The expected outcomes reproduce from the bytes in `vectors.json` alone, so a second implementation reaches the same verdicts without sharing code. aeoess ran them independently and got identical results, which was the point of writing them as bytes rather than prose.

On the wider picture in #243, the admission-controller machinery may well belong elsewhere, in SIG-Auth or SIG-Security, or in an existing project like sigstore policy-controller. I have no attachment to landing that here, and your scope instinct is right. If the SIG decides the broader admission-controller proposal belongs elsewhere, these vectors can move with that boundary; they do not depend on this repo owning the admission machinery. Where the reusable part could actually help this SIG is narrower and closer to the observability line: keeping a recorded authorization decision honest about what it attests, which is the same distinction coming up in the tracing work on #280.

I cannot make Thursday's meeting, so I wanted to put the short version in writing. Happy to follow whatever routing you prefer, and to move the vectors to wherever that boundary is worth making reproducible.

LiorLieberman:
Hi @Rul1an I am closing this PR for now based on discussion on the [issue](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/243) - feel free to comment there or join the meeting at one of the Thursday's. We also have a slack channel linked there if you wish to discuss this more. 

Thanks!

PR #305, opened 2026-06-18
fix: correct changelog filename and XAccessPolicy typo

david-martin:
## Summary
- Rename `CHANGELOG/0.x-CHANGELOG.md` to `CHANGELOG/0.1-CHANGELOG.md` to match the convention documented in RELEASE.md
- Fix "XAccessControl" typo to "XAccessPolicy"

## Test plan
- [ ] Verify renamed file renders correctly on the release branch


LiorLieberman (commented):
Thanks @david-martin !!
/lgtm
/approve

LiorLieberman:
TL;DR- I think the error is because you have a slash in the branch name.. can you try a different branch with a different branch name? 

More context:

the `post-agentic-net-push-controller-image` job is failing because Docker tags don't support slashes, and the build is trying to tag the image with your branch name.

While `hack/build/Makefile.common.in` attempts to replace slashes with dashes, Prow injects the branch name as an environment variable. And this likely overrides (I think) the `?=` assignment in the Makefile, bypassing the sanitization.

Alternatively - can you change `hack/build/Makefile.common.in` and add this line after declaring BASE_REF, which will substitute slashes with "-" regardless.

`override BASE_REF := $(subst /,-,$(BASE_REF))`

LiorLieberman:
Anything else beyond merging this @bexxmodd ?

bexxmodd:
changelog is named 0.x on purpose as we'll re-use the same file for every release before GA. So, let's fix the XAccessPolicy only here.

david-martin:
Reverted the rename — kept `0.x-CHANGELOG.md`. PR now only fixes the XAccessPolicy typo.

david-martin:
@bexxmodd  Should we update RELEASE.md to match? It currently says `CHANGELOG/X.Y-CHANGELOG.md` in a few places, e.g. [line 23](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/release-0.1/RELEASE.md?plain=1#L23).

LiorLieberman:
>changelog is named 0.x on purpose as we'll re-use the same file for every release before GA. So, let's fix the XAccessPolicy only here.

@bexxmodd shouldn't release-0.1 branch have that 0.1 instead of 0.x? and we keep 0.x in main?

bexxmodd:
> @bexxmodd Should we update RELEASE.md to match? It currently says CHANGELOG/X.Y-CHANGELOG.md in a few places, e.g. [line 23](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/release-0.1/RELEASE.md?plain=1#L23).

Yeah I'll need to update RELEASE.md anyways so I'll change that too.

> @bexxmodd shouldn't release-0.1 branch have that 0.1 instead of 0.x? and we keep 0.x in main?

there's no necessity for this. Just 0.x pre GA and 1.[0-9] for each minor release and that's it.

PR #309, opened 2026-06-25
add l7 params to xaccesspolicy

### LINKED ISSUE #307: Add AccessPolicy (L7) attribute matchers beyond source
More attribute matchers for the request like path/headers etc for in XAccessPolicy

**Why this is needed**:
We discussed this in length over oss meetings and gh discussion and had a consensus we are going to add more l7 attribute matchers. https://github.com/kubernetes-sigs/kube-agentic-networking/discussions/235#discussioncomment-16506423

/cc @david-martin

LiorLieberman:
/kind feature

This PR is a follow-up to the discussion in [Discussions #235](https://github.com/kubernetes-sigs/kube-agentic-networking/discussions/235#discussioncomment-16506423) and addresses [Issue #307](https://github.com/kubernetes-sigs/kube-agentic-networking/issues/307) by introducing L7 request attribute matchers to the XAccessPolicy.

There are two main things here;
- Adding additonal matchers (paths, methods, headers, hosts, ports) directly to AuthorizationRule for inline authorization.

- [This one likely need more discussion] Making AccessRule.Source optional to support allowing cluster endpoints i.e `tools/list` regardless of the the source. If Source remained required, users would have to enumerate every possible client workload identity just to apply an L7 header, path restriction and/or specific mcp method/tool.

Fixes #307 

```release-note
add additional request matchers to XAccessPolicy
```


david-martin:
/lgtm

The behaviour to make source optional is reasonable, a good usability change.
Methods AND semantics makes sense. OR for the others is correct since a request can't match two ports/paths/hosts simultaneously (guessing this is similar elsewhere in gateway api)

Need follow-on issues for the implementation/controller side of source-optional behaviour, and matchers?

PR #312, opened 2026-06-26
:memo: Update observability proposal evaluation to decision spans

evaline-ju:
**What type of PR is this?**

/kind documentation

With the implementation for tracing from agent to gateway for the reference implementation in #280 and follow-on discussion, it was decided that instead of advising spans to be present for every (access policy) rule evaluated, we should keep this to the rule that affected the decision to allow/deny. This PR updates the original observability proposal that #280 was implementing to help account for this 

```release-note
Proposal update
```
/cc @david-martin

### DIFF
--- a/docs/proposals/0033-Observability.md
+++ b/docs/proposals/0033-Observability.md
@@ -44,12 +44,14 @@ These can be considered for inclusion in an existing or new OpenTelemetry semant
 
 ### Per-Rule Evaluation Spans
 
-For authorization and guardrail checks, emit an `mcp.authorization.rule` or `mcp.guardrail.rule` child span for each rule evaluated within the parent span. Each child span records:
+For authorization and guardrail checks, implementations should emit a single `mcp.authorization.rule` or `mcp.guardrail.rule` child span identifying the rule that determined the outcome. Each span records:
 - `security_rule.name` — the rule identifier
 - `security_rule.match` — whether the rule matched (`true`/`false`)
-- `event.action` — the action the rule would take (`allow`/`deny`)
+- `event.action` — the action taken (`allow`/`deny`)
 
-Evaluation stops at the determining rule. For authorization, if no allow rule matches, emit an explicit `default-deny` rule span (`security_rule.name: default-deny`, `security_rule.match: true`, `event.action: deny`) to make the implicit default-deny behavior visible in traces.
+This span represents decisive attribution — which rule caused the outcome — rather than a full evaluation log. Implementations may additionally emit spans for rules that were evaluated but did not contribute to the decision, which can be useful for audit purposes.
+
+For authorization, if no explicit allow rule matches, implementations should emit an `mcp.authorization.rule` span with `security_rule.name: default-deny`, `security_rule.match: true`, and `event.action: deny` to make the implicit default-deny behavior visible in traces, rather than leaving the denial attributed only to the parent span.
 
 ## Retries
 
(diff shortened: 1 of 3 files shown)


docs/proposals/0033-Observability.md
```diff
@@ -44,12 +44,14 @@ These can be considered for inclusion in an existing or new OpenTelemetry semant
...
 - `security_rule.name` — the rule identifier
 - `security_rule.match` — whether the rule matched (`true`/`false`)
-- `event.action` — the action the rule would take (`allow`/`deny`)
+- `event.action` — the action taken (`allow`/`deny`)
 
-Evaluation stops at the determining rule. For authorization, if no allow rule matches, emit an explicit `default-deny` rule span (`security_rule.name: default-deny`, `security_rule.match: true`, `event.action: deny`) to make the implicit default-deny behavior visible in traces.
+This span represents decisive attribution — which rule caused the outcome — rather than a full evaluation log. Implementations may additionally emit spans for rules that were evaluated but did not contribute to the decision, which can be useful for audit purposes.
+
+For authorization, if no explicit allow rule matches, implementations should emit an `mcp.authorization.rule` span with `security_rule.name: default-deny`, `security_rule.match: true`, and `event.action: deny` to make the implicit default-deny behavior visible in traces, rather than leaving the denial attributed only to the parent span.
```
david-martin: for the avoidance of doubt/conflict, in case someone defines a rule with the name 'default-deny', what do you think about a separate span like `mcp.authorization.default-deny`, or something like `security_rule.implicit: true`

david-martin (commented):
/approve
The change makes sense given discussions on community calls.

Holding lgtm pending question inline

docs/proposals/0033-Observability.md
evaline-ju: thanks for the suggestion! I leaned toward the `security_rule.implicit` as likely more extensible for when our rules aren’t just "allow" rules and less tied to naming (potentially a user could write their own ‘default-deny’?) 

With the updates I then noticed `security_rule.match` was technically an ‘extension’ on the official Otel [security_rule attributes](https://opentelemetry.io/docs/specs/semconv/registry/attributes/security-rule/) that we hadn’t called out separately before, but also it would be less meaningful with the first-match-wins idea, so I’ve updated the proposal wording and example a bit. Happy to revise if you see this part differently

Rul1an:
The decisive-attribution split reads exactly right to me, and pairing it with "evaluation stops at the determining rule" draws the boundary cleanly.

One small precision on the optional fuller log, just so the wording stays unambiguous for a later reader. Since evaluation stops at the determining rule, `security_rule.match: false` is most honest when it strictly means "evaluated and did not match", which is the case for a rule that ran before the winning one. A rule ordered after the winner is never reached, so it has no match verdict to carry; reading the proposal as written, those simply get no span, which is right. If an implementation ever wants policy-coverage traces that also cover the rules the engine never reached, those read cleaner as a separate `skipped` / `not_evaluated` state than as `match: false`, so not-reached and evaluated-false stay distinguishable.

Mostly this just confirms the boundary the text already draws. Happy to suggest a one-line note if it is worth making explicit.

david-martin:
@gkhom @howardjohn fyi

gkhom (approved):
LGTM

evaline-ju:
seems like we keep hitting a timeout for golangci-lint, unrelated to these doc updates

LiorLieberman:
Thanks! I added lgtm but actually I prefer to add it in .golangci.yaml under run.timeout if you can

LiorLieberman (commented):
/lgtm

PR #315, opened 2026-07-08
l7 impl

### LINKED ISSUE #307: Add AccessPolicy (L7) attribute matchers beyond source

LiorLieberman:
**What type of PR is this?**
/kind feature

align impl with #309 
Fixes #307 

```release-note
support l7 params
```
/cc @david-martin @bexxmodd @gkhom

### DIFF
--- a/pkg/translator/accesspolicy.go
+++ b/pkg/translator/accesspolicy.go
@@ -180,7 +180,7 @@ func (t *Translator) mergeAllowPoliciesToRBAC(policies []*agenticv1alpha1.XAcces
 
 	for _, policy := range policies {
 		for _, rule := range policy.Spec.Rules {
-			policyName := rule.Name
+			policyName := fmt.Sprintf("%s-%s", policy.Name, rule.Name)
 			rbacPolicy := t.translateAccessRuleToRBACPolicy(policy.Namespace, rule)
 			if rbacPolicy == nil {
 				continue
--- a/pkg/translator/accesspolicy.go
+++ b/pkg/translator/accesspolicy.go
@@ -181,49 +181,9 @@ func (t *Translator) mergeAllowPoliciesToRBAC(policies []*agenticv1alpha1.XAcces
 	for _, policy := range policies {
 		for _, rule := range policy.Spec.Rules {
 			policyName := rule.Name
-			source := t.ruleSourceToPrincipalName(policy.Namespace, rule.Source)
-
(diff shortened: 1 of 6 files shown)


pkg/translator/accesspolicy.go
```diff
@@ -393,6 +315,265 @@ func (t *Translator) translateAccessPolicyToRBAC(accessPolicy *agenticv1alpha1.X
...
+				},
+			})
+		}
+	}
+
+	// 4. Ports
+	if len(auth.Ports) > 0 {
+		var portRules []*rbacconfigv3.Permission
+		for _, port := range auth.Ports {
+			if port < 0 {
```
david-martin, line 482: <= 0 ?

david-martin (commented):
generally lgtm (most discussion was in the issue).

Thoughts on going a little deeper in the unit tests to verify the envoy matcher content that's output?

LiorLieberman:
>Thoughts on going a little deeper in the unit tests to verify the envoy matcher content that's output?

Not sure I follow what you mean?

david-martin:
> > Thoughts on going a little deeper in the unit tests to verify the envoy matcher content that's output?
> 
> Not sure I follow what you mean?

Ah, sorry. I mean to check the rule contents match what's expected after checking the number of rules is correct.
Small thing to boost test confidence.

pkg/translator/accesspolicy.go
LiorLieberman: I mean, 0 is also not a port that folks would use cause its reserved but its valid one, much like 1,2,3 etc. any reason we should consider 0 here in this block as a case to continue and not process the rest?

david-martin: i think it's fine actually.
Using 0 would mean a rule with port 0 gets adding in envoy, which wouldn't match anything in practice.

LiorLieberman:
adding authorino in-tree cause we sometimes have transient errors trying to download it

FYi @guicassolato @david-martin

david-martin:
/lgtm

Thanks for this!

PR #319, opened 2026-07-29
feat: add Programmed status condition to XAccessPolicy (#314)

### LINKED ISSUE #314: `Programmed` status condition in the XAccessPolicy

Add a [`Programmed`](https://gateway-api.sigs.k8s.io/geps/gep-713/#status-conditions) status condition to the XAccessPolicy CRD.

**Why this is needed**:

To allow implementations to signal when the target resources of a policy have been configured for enforcement of the policy. This is not a guarantee that the policy is being enforced – which is often off-limits for the controllers to tell – but it allows differentiating from an `Accepted` status condition which basically means the policy is semantically valid though not necessarily configured for enforcement yet.

Conformance tests should _test for the presence_ of this new condition (specified as a SHOULD in GEP-713), as well as build on the condition to _decide when to ready to send traffic_ to the targets of the policy.

Policy resources whose `Programmed` status condition is unset or set to `False` should **not** be considered ready for enforcement. Therefore, their targets should be assumed not protected by the rules specified in the policy and _possibly_ not ready to receive traffic safely.

Dasmat13:
Fixes #314

### What this PR does / why we need it
This PR implements GEP-713 standard [`Programmed` status condition](https://gateway-api.sigs.k8s.io/geps/gep-713/#status-conditions) support on `XAccessPolicy` resources, following the implementation design proposed by @vibhor-5 in #314.

- Adds `PolicyConditionProgrammed` (`Programmed`) and `PolicyReasonProgrammed` (`Programmed`) constants to `v1alpha1` and `v0alpha0` API packages.
- Updates `updateAccessPolicyStatus` in `pkg/controller/accesspolicy.go` to populate both `Accepted` and `Programmed` conditions on ancestor status records.
- Adds `IsXAccessPolicyProgrammed` helper in `pkg/helpers/accesspolicy.go` and `api/v0alpha0/helpers/accesspolicy.go`.
- Adds unit test coverage in `pkg/controller/accesspolicy_test.go`, `pkg/helpers/accesspolicy_test.go`, and `api/v0alpha0/helpers/accesspolicy_test.go`.
- Updates `conformance/tests/xaccesspolicy-allow-accepted.go` to wait for `Accepted` and `Programmed` conditions.

*Credit & thanks to @vibhor-5 for the initial implementation plan proposal on #314!*

### Release Note
```release-note
Added standard GEP-713 `Programmed` status condition to `XAccessPolicy` ancestor status.
```

### DIFF
--- a/api/v0alpha0/accesspolicy_types.go
+++ b/api/v0alpha0/accesspolicy_types.go
@@ -196,9 +196,26 @@ const (
 	//
 	PolicyConditionAccepted gwapiv1.PolicyConditionType = "Accepted"
 
+	// PolicyConditionProgrammed indicates whether the policy has been programmed by the controller into the data plane.
+	//
+	// Possible reasons for this condition to be True are:
+	//
+	// * "Programmed"
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "LimitPerTargetExceeded"
+	// * "InvalidCEL"
+	//
+	PolicyConditionProgrammed gwapiv1.PolicyConditionType = "Programmed"
+
 	// This reason is used with the "Accepted" condition when the policy
 	// has been accepted by the controller.
 	PolicyReasonAccepted gwapiv1.PolicyConditionReason = "Accepted"
+
+	// This reason is used with the "Programmed" condition when the policy
+	// has been programmed by the controller into the data plane.
+	PolicyReasonProgrammed gwapiv1.PolicyConditionReason = "Programmed"
 
 	// This reason is used with the "Accepted" condition when the policy
 	// was rejected because the maximum number of policies per target was exceeded.
--- a/api/v0alpha0/accesspolicy_types.go
+++ b/api/v0alpha0/accesspolicy_types.go
@@ -217,6 +217,10 @@ const (
 	// has been programmed by the controller into the data plane.
 	PolicyReasonProgrammed gwapiv1.PolicyConditionReason = "Programmed"
 
+	// This reason is used with the "Programmed" condition when the policy
+	// is pending programming by the controller into the data plane.
+	PolicyReasonPending gwapiv1.PolicyConditionReason = "Pending"
+
 	// This reason is used with the "Accepted" condition when the policy
 	// was rejected because the maximum number of policies per target was exceeded.
 	PolicyLimitPerTargetExceeded gwapiv1.PolicyConditionReason = "LimitPerTargetExceeded"
--- a/api/v1alpha1/accesspolicy_types.go
+++ b/api/v1alpha1/accesspolicy_types.go
@@ -366,6 +366,10 @@ const (
 	// has been programmed by the controller into the data plane.
 	PolicyReasonProgrammed gwapiv1.PolicyConditionReason = "Programmed"
 
+	// This reason is used with the "Programmed" condition when the policy
+	// is pending programming by the controller into the data plane.
+	PolicyReasonPending gwapiv1.PolicyConditionReason = "Pending"
+
 	// This reason is used with the "Accepted" condition when the policy
 	// was rejected because the maximum number of policies per target was exceeded.
 	PolicyLimitPerTargetExceeded gwapiv1.PolicyConditionReason = "LimitPerTargetExceeded"
--- a/api/v1alpha1/accesspolicy_types.go
+++ b/api/v1alpha1/accesspolicy_types.go
@@ -345,9 +345,26 @@ const (
 	//
 	PolicyConditionAccepted gwapiv1.PolicyConditionType = "Accepted"
 
+	// PolicyConditionProgrammed indicates whether the policy has been programmed by the controller into the data plane.
+	//
+	// Possible reasons for this condition to be True are:
+	//
+	// * "Programmed"
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "LimitPerTargetExceeded"
+	// * "InvalidCEL"
+	//
+	PolicyConditionProgrammed gwapiv1.PolicyConditionType = "Programmed"
+
 	// This reason is used with the "Accepted" condition when the policy
 	// has been accepted by the controller.
 	PolicyReasonAccepted gwapiv1.PolicyConditionReason = "Accepted"
+
+	// This reason is used with the "Programmed" condition when the policy
+	// has been programmed by the controller into the data plane.
+	PolicyReasonProgrammed gwapiv1.PolicyConditionReason = "Programmed"
 
 	// This reason is used with the "Accepted" condition when the policy
 	// was rejected because the maximum number of policies per target was exceeded.
--- a/conformance/tests/accesspolicy-accepted.go
+++ b/conformance/tests/accesspolicy-accepted.go
@@ -55,11 +55,12 @@ var XAccessPolicyAccepted = suite.ConformanceTest{
 			}
 
 			accepted := helpers.IsXAccessPolicyAccepted(policy)
-			t.Logf("XAccessPolicy %s Accepted condition: %v", policyName.Name, accepted)
+			programmed := helpers.IsXAccessPolicyProgrammed(policy)
+			t.Logf("XAccessPolicy %s Accepted: %v, Programmed: %v", policyName.Name, accepted, programmed)
 
-			return accepted, nil
+			return accepted && programmed, nil
 		})
-		require.NoError(t, err, "timed out waiting for XAccessPolicy to be accepted")
+		require.NoError(t, err, "timed out waiting for XAccessPolicy to be accepted and programmed")
 	},
 }
 
--- a/conformance/tests/accesspolicy-accepted.go
+++ b/conformance/tests/accesspolicy-accepted.go
@@ -55,12 +55,11 @@ var XAccessPolicyAccepted = suite.ConformanceTest{
 			}
 
 			accepted := helpers.IsXAccessPolicyAccepted(policy)
-			programmed := helpers.IsXAccessPolicyProgrammed(policy)
-			t.Logf("XAccessPolicy %s Accepted: %v, Programmed: %v", policyName.Name, accepted, programmed)
+			t.Logf("XAccessPolicy %s Accepted condition: %v", policyName.Name, accepted)
 
-			return accepted && programmed, nil
+			return accepted, nil
 		})
-		require.NoError(t, err, "timed out waiting for XAccessPolicy to be accepted and programmed")
+		require.NoError(t, err, "timed out waiting for XAccessPolicy to be accepted")
 	},
 }
 
--- a/conformance/tests/xaccesspolicy-allow-accepted.go
+++ b/conformance/tests/xaccesspolicy-allow-accepted.go
@@ -55,12 +55,11 @@ var XAccessPolicyAllowAccepted = suite.ConformanceTest{
 			}
 
 			accepted := helpers.IsXAccessPolicyAccepted(policy)
-			programmed := helpers.IsXAccessPolicyProgrammed(policy)
-			t.Logf("XAccessPolicy %s Accepted: %v, Programmed: %v", policyName.Name, accepted, programmed)
+			t.Logf("XAccessPolicy %s Accepted condition: %v", policyName.Name, accepted)
 
-			return accepted && programmed, nil
+			return accepted, nil
 		})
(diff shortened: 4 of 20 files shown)


api/v0alpha0/accesspolicy_types.go
```diff
@@ -196,10 +196,27 @@ const (
...
+	PolicyConditionProgrammed gwapiv1.PolicyConditionType = "Programmed"
+
 	// This reason is used with the "Accepted" condition when the policy
 	// has been accepted by the controller.
 	PolicyReasonAccepted gwapiv1.PolicyConditionReason = "Accepted"
 
+	// This reason is used with the "Programmed" condition when the policy
+	// has been programmed by the controller into the data plane.
+	PolicyReasonProgrammed gwapiv1.PolicyConditionReason = "Programmed"
+
```
vibhor-5, line 219: I think a new `PolicyConditionReason` Pending can be useful here and maybe some others as well

api/v1alpha1/accesspolicy_types.go
```diff
@@ -345,10 +345,27 @@ const (
...
+	//
+	PolicyConditionProgrammed gwapiv1.PolicyConditionType = "Programmed"
+
 	// This reason is used with the "Accepted" condition when the policy
 	// has been accepted by the controller.
 	PolicyReasonAccepted gwapiv1.PolicyConditionReason = "Accepted"
 
+	// This reason is used with the "Programmed" condition when the policy
+	// has been programmed by the controller into the data plane.
+	PolicyReasonProgrammed gwapiv1.PolicyConditionReason = "Programmed"
```
vibhor-5, line 367: Same here as `api/v0alpha0/accesspolicy_types.go`

conformance/tests/xaccesspolicy-allow-accepted.go
vibhor-5, line 1: The following tests currently only check `IsXAccessPolicyAccepted(policy)` and must be updated to also check `IsXAccessPolicyProgrammed(policy)` before sending traffic:
  • conformance/tests/xaccesspolicy-base-protocol-match.go
  • conformance/tests/xaccesspolicy-base-protocol-skip.go
  • conformance/tests/xaccesspolicy-evaluation-logic.go
  • conformance/tests/xaccesspolicy-extauth-accepted.go
  • conformance/tests/xaccesspolicy-extauth-limit.go
  • conformance/tests/xaccesspolicy-sa-matching.go
  • conformance/tests/xaccesspolicy-spiffe-matching.go
  • conformance/tests/accesspolicy-accepted.go

pkg/controller/accesspolicy.go
vibhor-5, line 196: check for programmed as well

vibhor-5, line 141: check for programmed as well

vibhor-5, line 1: Currently the reference controller immediately sets `Programmed=True`at the exact same time as `Accepted=True`. While this is an acceptable implementation detail for a reference controller (simulating zero-latency dataplane updates), it bypasses the asynchronous intent of the feature. We might want to decouple these later if we ever simulate real dataplane latency, or do it asynchronously itself in reference controller for better standard practices

vibhor-5 (changes requested):
I might have missed some things, but yeah main concern is to change places only checking for only `Accepted` to also check for `Programmed` where needed. And update the conformance tests

api/v0alpha0/accesspolicy_types.go
Dasmat13: I'll add PolicyReasonPending to v0alpha0 for when the policy is pending programming

api/v1alpha1/accesspolicy_types.go
Dasmat13: ok  adding PolicyReasonPending to v1alpha1 as well

conformance/tests/xaccesspolicy-allow-accepted.go
Dasmat13: Thanks for listing these out! I'll update all 8 conformance tests to check IsXAccessPolicyProgrammed(policy) before sending traffic

pkg/controller/accesspolicy.go
Dasmat13: I'll update this to check programmedChanged as well

Dasmat13: Agreed, I'll update isAccepted logic to also check IsXAccessPolicyProgrammed

Dasmat13: for now setting both together works fine for zero latency reference ctrl, but we can decouple them later when doing real async dataplane sync

Dasmat13:
Hi @LiorLieberman @guicassolato @vibhor-5 👋

Friendly check-in on this PR. All feedback from previous comments has been addressed, and tests/linting are passing cleanly. 

Could a maintainer please add `/ok-to-test` when you get a chance? Thanks!

Dasmat13:
Hi @vibhor-5, quick ping — I addressed your review feedback. Could you take another look when you get a chance? Thanks

vibhor-5 (changes requested):
Hey @Dasmat13 Thanks for the changes just couple more minor changes in pkg/controller/accesspolicy.go:
  1. Decouple Accepted and Programmed
  Right now, updateAccessPolicyStatus sets Programmed=True at the exact same time as Accepted. Since the goal of the    issue is to differentiate between a valid policy (Accepted) and one that's actually enforced in the data  (Programmed), we shouldn't set them together during the validation phase.
  Instead, we should initialize Programmed to False ,using the PolicyReasonPending reason you added and when the policy is accepted,only then it should transition to True downstream once the data plane translation/sync actually succeeds.
  
  2. Gateway queueing logic
  In enqueueGatewaysForAccessPolicy, you updated shouldEnqueueGW to require isProgrammed to be true. If we decouple the conditions as mentioned above, the Gateway will never be enqueued because Programmed will initially be False; which creates a chicken-and-egg problem since the Gateway sync is what actually does the programming
  Let's revert that line back to isAccepted || isDeleting.

Otherwise LGTM.

Dasmat13:
Thanks @vibhor-5 Updated as requested:

Decoupled Programmed from Accepted in updateAccessPolicyStatus so it starts as ConditionFalse with PolicyReasonPending.
Reverted shouldEnqueueGW back to isAccepted || isDeleting.
Updated unit tests accordingly.
Just pushed the update

pkg/controller/accesspolicy.go
```diff
@@ -352,7 +370,8 @@ func (c *Controller) updateAccessPolicyStatus(ctx context.Context, policy *agent
 			ancestorStatus = &policyCopy.Status.Ancestors[len(policyCopy.Status.Ancestors)-1]
 		}
 
-		meta.SetStatusCondition(&ancestorStatus.Conditions, newCondition)
+		meta.SetStatusCondition(&ancestorStatus.Conditions, acceptedCondition)
+		meta.SetStatusCondition(&ancestorStatus.Conditions, programmedCondition)
```
guicassolato, line 374: Here `Programmed` is set during the acceptance phase, not after xDS programming. Both `Accepted` and `Programmed` conditions are set simultaneously. When a policy is accepted, `Programmed` is set to `False/Pending`; when rejected, `Programmed` is set to `False/<rejection-reason>`.

I couldn't find anywhere in code path that sets `Programmed` to `True`. The `PolicyReasonProgrammed` constant is defined but never used in the controller, right?

Per [GEP-713](https://gateway-api.sigs.k8s.io/geps/gep-713/#status-conditions), `Programmed` should transition to `True` when the policy has actually been configured in the data plane. IMO, the natural place for this is after the xDS push succeeds at [`controller.go:478`](https://github.com/Dasmat13/kube-agentic-networking/tree/feat/xaccesspolicy-programmed-condition#L478). After that call to `UpdateXDSServer(…, resources)` returns successfully, the controller knows the gateway's data plane has been programmed with the policies, and should update the policies' `Programmed` condition per-ancestor for that gateway.

conformance/tests/accesspolicy-accepted.go
```diff
@@ -55,11 +55,12 @@ var XAccessPolicyAccepted = suite.ConformanceTest{
 			}
 
 			accepted := helpers.IsXAccessPolicyAccepted(policy)
-			t.Logf("XAccessPolicy %s Accepted condition: %v", policyName.Name, accepted)
+			programmed := helpers.IsXAccessPolicyProgrammed(policy)
+			t.Logf("XAccessPolicy %s Accepted: %v, Programmed: %v", policyName.Name, accepted, programmed)
 
-			return accepted, nil
+			return accepted && programmed, nil
```
guicassolato, line 61: Why coupling `Programmed` into existing `Accepted` conformance tests instead of being a separate test?

Every conformance test that previously waited for `Accepted` now also waits for `Programmed`. This means a test named "accepted" now tests two conditions, muddying what it verifies. Since `Programmed` is never set to `True` (see comment below), these tests should all timeout (unless I'm missing something about the test harness).

We need a dedicated conformance test that exercises the `Programmed` lifecycle independently — e.g., verifying the transition from `Pending` to `Programmed` after data plane configuration.

pkg/controller/accesspolicy.go
```diff
@@ -194,8 +194,9 @@ func hasAccessPolicyChanged(oldPolicy, newPolicy *agenticv1alpha1.XAccessPolicy)
 	specChanged := newPolicy.Generation != oldPolicy.Generation || !reflect.DeepEqual(newPolicy.Annotations, oldPolicy.Annotations)
 	deletionTimestampChanged := newPolicy.DeletionTimestamp != oldPolicy.DeletionTimestamp
 	acceptanceChanged := helpersv1alpha1.IsXAccessPolicyAccepted(newPolicy) != helpersv1alpha1.IsXAccessPolicyAccepted(oldPolicy)
+	programmedChanged := helpersv1alpha1.IsXAccessPolicyProgrammed(newPolicy) != helpersv1alpha1.IsXAccessPolicyProgrammed(oldPolicy)
 
-	return specChanged || deletionTimestampChanged || acceptanceChanged
+	return specChanged || deletionTimestampChanged || acceptanceChanged || programmedChanged
```
guicassolato, line 199: When the `Programmed` condition is set on a policy (via the status update), the informer fires an update event. This func then sees the `Programmed` condition changed and triggers a new reconciliation, which re-enqueues gateways via `enqueueGatewaysForAccessPolicy`. Although this may be useful if `Programmed` was set outside the gateway reconciliation loop (e.g., after xDS push), in the same status update as `Accepted`, it creates an extra unnecessary reconciliation cycle, only to reflect the same information one could already infer from `Accepted`.

guicassolato (changes requested):
I have to echo one of @vibhor-5's [previous concerns](https://github.com/kubernetes-sigs/kube-agentic-networking/pull/319#pullrequestreview-4893629123) about coupling between `Accepted` and `Programmed`.

I think the outline of what we're looking for is more or less the following:
1. During `syncAccessPolicy` / `isPolicyUnderTargetLimit`: set only the `Accepted` condition (as it does today on `main`). Optionally set `Programmed=False/Pending` at this point as an initial state.
2. In `syncGateway`, after [`c.xdsServer.UpdateXDSServer`](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/5bf62b6a9ca750f95ddb27cd5d66c34c57cb9e7d/pkg/controller/controller.go#L478) succeeds: iterate over the policies that were included in the xDS translation and set `Programmed=True` for the ancestor corresponding to that gateway. If the xDS push fails, set `Programmed=False` with an appropriate reason. (Watch out not to loop reconciliation calls after this event.)
3. Separately, add a dedicated `xaccesspolicy-programmed` conformance test that specifically verifies the `Programmed` lifecycle.
Don't modify existing `Accepted` conformance tests to also assert on `Programmed`.

conformance/tests/xaccesspolicy-allow-accepted.go
guicassolato: I have my doubts we want `conformance/tests/accesspolicy-accepted.go` asserting for `Programmed` or rather a separate conformance test for this new status type.

vibhor-5:
Hey @Dasmat13, will you be able to continue working on the pr ? , if not lmk I can take this up and continue forward.

Dasmat13:
@vibhor-5 Sorry for the delay! Yes, I’ll continue working on the PR and should be able to complete it in the next few days. Thanks for checking in!

Dasmat13:
@guicassolato updated this based on your review, should be ready for another look

PR #323, opened 2026-08-25
fix: set HTTPRoute status conditions when LoadBalancer is pending

### LINKED ISSUE #322: HTTPRoute status conditions not set when LoadBalancer address is pending
## What happened

When `EnsureProxyExist` returns an error (e.g., LoadBalancer IP not yet assigned on a kind cluster without MetalLB), `syncGateway` returns early and never calls `TranslateGatewayToXDS` or `updateHTTPRouteStatuses`.

This means HTTPRoute `Accepted` and `ResolvedRefs` conditions are never set, even though they are logically independent of whether the proxy has an address. Per the Gateway API spec, route acceptance should be determined as soon as the controller can validate the route.

## Steps to reproduce

1. Create a kind cluster (no MetalLB)
2. Install Gateway API CRDs + kube-agentic-networking
3. Create a Gateway with `gatewayClassName: kube-agentic-networking`
4. Create an HTTPRoute referencing the Gateway
5. `kubectl get httproute <name> -o yaml`

**Expected:** `status.parents[].conditions` includes `Accepted=True`, `ResolvedRefs=True`
**Actual:** No `status` section at all

The Envoy proxy Deployment is created and Running, but the LoadBalancer Service stays `<pending>`. The controller retries indefinitely at the `EnsureProxyExist` step and never reaches the HTTPRoute status update.

## Root cause

In `pkg/controller/controller.go`, `syncGateway` lines 441-445:

```go
proxyAddress, err := rm.EnsureProxyExist(ctx)
if err != nil {
    updateErr := updateGatewayStatus(ctx, c, gateway, newGW, nil, err)
    return errors.Join(err, updateErr)
}
```

`TranslateGatewayToXDS` (which computes HTTPRoute statuses) and `updateHTTPRouteStatuses` are only called after this block, so they are skipped on error.

## Suggested fix

Before the early return, call `TranslateGatewayToXDS` and `updateHTTPRouteStatuses` so HTTPRoute conditions are set regardless of proxy readiness:

```go
proxyAddress, ensureProxyErr := rm.EnsureProxyExist(ctx)
if ensureProxyErr != nil {
    if c.translator != nil {
        if _, _, httpRouteStatuses, _, translateErr := c.translator.TranslateGatewayToXDS(ctx, gateway); translateErr == nil {
            _ = c.updateHTTPRouteStatuses(ctx, httpRouteStatuses)
        }
    }
    updateErr := updateGatewayStatus(ctx, c, gateway, newGW, nil, ensureProxyErr)
    return errors.Join(ensureProxyErr, updateErr)
}
```

/kind bug

matzew:
When EnsureProxyExist fails (e.g. LoadBalancer address not yet assigned), syncGateway returns early before calling TranslateGatewayToXDS or updateHTTPRouteStatuses. This leaves HTTPRoute Accepted/ResolvedRefs conditions unset even though route validation is independent of proxy readiness.

Call TranslateGatewayToXDS and updateHTTPRouteStatuses before the early return so HTTPRoute status is populated regardless of infrastructure state.

Fixes #322

**What type of PR is this?**
/kind bug

It was found when looking at https://github.com/kubernetes-sigs/mcp-lifecycle-operator/pull/355

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -438,10 +438,15 @@ func (c *Controller) syncGateway(ctx context.Context, key string) error {
 
 	// Ensure Envoy proxy deployment and service exist.
 	rm := envoy.NewResourceManager(c.core.client, gateway, c.envoyImage, c.agenticIdentityTrustDomain)
-	proxyAddress, err := rm.EnsureProxyExist(ctx)
-	if err != nil {
-		updateErr := updateGatewayStatus(ctx, c, gateway, newGW, nil, err)
-		return errors.Join(err, updateErr)
+	proxyAddress, ensureProxyErr := rm.EnsureProxyExist(ctx)
+	if ensureProxyErr != nil {
+		if c.translator != nil {
+			if _, _, httpRouteStatuses, _, translateErr := c.translator.TranslateGatewayToXDS(ctx, gateway); translateErr == nil {
+				_ = c.updateHTTPRouteStatuses(ctx, httpRouteStatuses)
+			}
+		}
+		updateErr := updateGatewayStatus(ctx, c, gateway, newGW, nil, ensureProxyErr)
+		return errors.Join(ensureProxyErr, updateErr)
 	}
 
 	// TODO: Add support for IPv6?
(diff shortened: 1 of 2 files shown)


LiorLieberman:
Sweet, thank you @matzew !
/approve
/lgtm

